PR #49, opened 2021-08-19
Enable golangci-lint

### LINKED ISSUE #22: Add more linting
More than just `go vet`, at least!

benjaminjkraft:
## Summary:
It would be nice to have some linting beyond `go vet`!  Now we do.  I
started by copying the config from Khan/webapp.  I did remove a couple
of staticcheck checks that I didn't feel were useful.  (Note also that
exportloopref is the replacement for scopelint in newer golangci-lint.)

Included are all the needed lint fixes; most are stylistic but the
changes in the example are a (minor) bugfix.

Fixes #22.
Issue: https://github.com/Khan/genqlient/issues/22

## Test plan:
make check


aberkan (approved):
Watching code pass more linters is pretty

Makefile
```diff
@@ -3,6 +3,7 @@ example:
 	go run ./example/cmd/example/main.go
 
 check:
+	golangci-lint run ./..
```
dnerdy, line 6: Is `make check` meant to be the primary way to run tests in development? If so, it would be nice to update the README to mention this.

Also, do you have a separate `golangci-lint` binary installed for this project? `~/webapp/genfiles/go/bin/golangci-lint` is the binary on my path, and the config for this project seems to be incompatible. This is the error:
```
golangci-lint run ./..
WARN Gocritic check "typeDefFirst" was explicitly disabled via config. However, as this checkis disabled by default, there is no need to explicitly disable it via config.
ERRO Invalid gocritic settings: validation failed: disabled checker typeDefFirst doesn't exist, all existing checkers: [appendassign appendcombine argorder assignop badcall badcond badregexp boolexprsimplify builtinshadow captlocal caseorder codegencomment commentedoutcode commentedoutimport commentformatting defaultcaseorder deferunlambda deprecatedcomment docstub duparg dupbranchbody dupcase dupimport dupsubexpr elseif emptyfallthrough emptystringtest equalfold evalorder exitafterdefer filepathjoin flagderef flagname hexliteral hugeparam ifelsechain importshadow indexalloc initclause mapkey methodexprcall nestingreduce newderef nilvalreturn octalliteral offby1 paramtypecombine ptrtorefparam rangeexprcopy rangevalcopy regexpmust regexppattern regexpsimplify ruleguard singlecaseswitch sloppylen sloppyreassign sloppytypeassert sortslice sqlquery stringxbytes switchtrue truncatecmp typeassertchain typeswitchvar typeunparen underef unlabelstmt unlambda unnamedresult unnecessaryblock unnecessarydefer unslice valswap weakcond whynolint wrapperfunc yodastyleexpr]
make: *** [check] Error 3
```

dnerdy: The above error is from #50, actually. But this PR also has an error in my environment.
```
golangci-lint run ./..
ERRO Running error: unknown linters: 'durationcheck', run 'golangci-lint linters' to see the list of supported linters
make: *** [check] Error 3
```

benjaminjkraft: Sure, I can mention it in the readme!  In practice most of the time I just run one or both of the commands manually.

And yes, this kinda assumes you have golangci-lint 1.42, which is not what Khan folks are likely to have.  I'll see if I can find a better way to run it.  It's surprisingly annoying to say "just run this command at this version" without updating the `go.mod`, but I can just put it in go.mod.

dnerdy (approved):
Thanks for pinning the linter version. The update works well!

Makefile
csilvers: I think that's the big feature add of go 1.17: `go run url@version`!

benjaminjkraft: Yep!  Even in 1.16 you can do `go install url@version` with similar behavior, but I figure I probably don't want to assume 1.16 quite yet, since in practice Khan folk don't have it!

Issue #24, opened 2021-04-22
Write tests for graphql package


benjaminjkraft:
They don't need to be super detailed -- there's not a lot of code -- but we should really have *something*!

PR #50, opened 2021-08-19
Add integration tests against a gqlgen server

### LINKED ISSUE #21: Add integration tests
We could add some integration tests against common GraphQL servers, to make sure we actually match what they expect.  Unclear how much of an issue this is in practice; the HTTP parts of GraphQL are a bit underspecified, but also fairly simple.

benjaminjkraft:
## Summary:
We have lots of tests covering codegen, but not a lot that actually run
the code.  For things where all we do is generate types, that's (mostly)
fine (especially now that we actually build the code), but as we
generate more nontrivial non-type code we need to actually run it.

So I wrote some integration tests that spin up a little gqlgen
server, and make calls to it; we can add more over time especially as
the JSON marshalling logic gets complex (to support fragments).
They're more work to write than the snapshot tests, but of course they
can test a lot more.

In addition to gqlgen, I pulled in testify assert/require, because I
really wanted to be able to use assert.Equal and such for these.  I
didn't bother converting existing tests, although I assume they will
become useful elsewhere in time.  Both gqlgen and testify are of course
only used in tests.

Fixes #21 and #24.
Issue: https://github.com/Khan/genqlient/issues/21
Issue: https://github.com/Khan/genqlient/issues/24

## Test plan:
make check


example/genqlient.yaml
```diff
@@ -1,11 +1,11 @@
 # These are the defaults, and are just included to be explicit.
 package: example
 schema: schema.graphql
-queries:
+operations:
```
dnerdy, line 4: I notice that generation still works when this key is `queries` since `genqlient.graphql` is the default value for `operations`. What do you think about using `yaml.UnmarshalStrict` to detect invalid keys in the configuration?

internal/integration/integration_test.go
```diff
@@ -0,0 +1,66 @@
+// Package integration contains genqlient's integration tests, which run
+// against a real server.
+//
+// These are especially important for cases where we generate nontrivial logic,
+// such as JSON-unmarshaling.
```
aberkan, line 5: You should add a comment pointing about where the test server code is.  Otherwise the magic test data appears out of nowhere.

aberkan (approved):

dnerdy (approved):
Nice! I appreciate the generation tests, and the test server will most certainly be helpful. I left a comment about perhaps performing some additional validation on the config, but that isn't blocking.

example/genqlient.yaml
benjaminjkraft: Yeah, that's probably a good idea.  This actually confused me for a while, because I copied this config for the integration tests, but actually wanted to change the value, and of course it didn't work.  I'll update it in a separate diff.

PR #51, opened 2021-08-19
Big refactor to separate operation-traversal from code-generation

benjaminjkraft:
## Summary:
I've felt for a while that types.go is way too confusing, and as I
started to implement some of the more complex cases of generating
interface-types, the cracks were really starting to show.  Luckily, I
also finally realized how to fix it: we need to separate the process of
traversing the GraphQL operation and schema to decide what types to
generate from the process of actually generating those types.  This
requires an extra set of intermediate data structures, but I think it
makes things quite a lot easier to understand -- and, importantly, it
means that the code-generation doesn't need to go in the order we
traverse the query/schema.

So in this commit, I did that huge refactor.  It's probably best to just
review types.go and traverse.go as if they were new; the old code was
quite hard to understand and the new code will hopefully make a lot more
sense.  (And to that end, review comments about what could be organized
better or needs more documentation are very much in order, even for code
that is mostly unchanged.)

This does introduce one bug, sort of, which is that rather than
generating broken code for list-of-interface fields, we generate no code
at all.  (A TODO in unmarshal.go describes why.)  I'll fix this when I
add support for those fields.  (It's all behind the AllowBrokenFeatures
flag, anyway.)  Otherwise, the only changes to generated code are that a
few methods are ordered differently, because we now generate the
implements-interface methods with the interface, rather than the
implementations, as it's much simpler that way.  (In GraphQL, unlike Go,
we know the list of all possible implementations of each interface, so
this is possible.)

## Test plan:
golangci-lint run ./... && go test ./...


dnerdy (approved):
I think having an intermediate data structure makes a lot of sense; I found it pretty easy to follow the new code. The feedback I have is pretty superficial:

- The file name `traverse.go` made sense once I understood what the code was doing, but it wasn't clear on the initial read that that was the file I needed to go to. Most of the functions in the file have a name that starts with "convert". Naming the file `convert.go` or something similar could help with discovery.
- Similarly, `comments.go` is related to genqlient directives, which happen to be powered by comments, but it wasn't clear to me at first that that's the place I should look for the directive code.
- Finally (and not at all related to this diff), I was surprised to see that template execution uses `runtime.Caller(0)` to locate a template. I think this only works when using `go run`; i.e. this would break if the user only has is a genqlient executable. Using `//go:embed` would work in both cases: https://pkg.go.dev/embed.

generate/traverse.go
```diff
@@ -0,0 +1,357 @@
...
+// calls to typeName on any child types.
+func (g *generator) typeName(prefix string, typ *ast.Definition) (name, nextPrefix string) {
+	typeGoName := upperFirst(typ.Name)
+	if typ.Kind == ast.Enum || typ.Kind == ast.InputObject {
+		// STOPSHIP: hopefully inline this
+		// If we're an enum or an input-object, there is only one type we
+		// will ever possibly generate for this type, so we don't need any
+		// of the qualifiers.  This is especially helpful because the
+		// caller is very likely to need to reference these types in their
+		// code.
```
dnerdy, line 92: I didn't see an obvious alternative place to put this. Is this comment suggesting the code be inlined in `convertField` (and ignore the place typeName is used in `convertDefinition` since enums and input objects aren't valid types to use with abstract fields)?

generate/types.go
```diff
@@ -1,412 +1,188 @@
...
+type goType interface {
+	// WriteDefinition writes the code for this type into the given io.Writer.
+	//
+	// TODO(benkraft): Some of the implementations might now benefit from being
+	// converted to templates.
+	WriteDefinition(io.Writer, *generator) error
+
+	// Reference returns the Go name of this type, e.g. []*MyStruct, and may be
+	// used to refer to it in Go code.
+	Reference() string
```
dnerdy, line 24: What do you think about naming this "TypeName" instead of "Reference". "Reference" sounds like this should always be a pointer.

benjaminjkraft:
>     * The file name `traverse.go` ...
>     * Similarly, `comments.go` ...

Good ideas, will do!

>     * Finally (and not at all related to this diff), I was surprised to see that template execution uses `runtime.Caller(0)` to locate a template. I think this only works when using `go run`; i.e. this would break if the user only has is a genqlient executable. Using `//go:embed` would work in both cases: https://pkg.go.dev/embed.
I believe it works as long as the executable was compiled on the same machine (and the genqlient checkout used to do so has not been moved, etc.) although I haven't tried that.  I agree embedding would be a better long-term solution, but we need to either use a third-party package or wait until we drop support for Go 1.13-1.15 (and we definitely can't drop 1.15 yet); see #9.

generate/traverse.go
benjaminjkraft: Whoops, I meant to remove this before landing, but apparently I only removed it in the next diff!  Let me see if I can backport the changes from there.

generate/types.go
benjaminjkraft: The problem is it's not always a type-name either since it might be `[]*MyStruct`.  But maybe that's at least closer.

Issue #8, opened 2021-04-22
Handle interfaces, unions, and fragments


benjaminjkraft:
See DESIGN.md for more on the plan.

PR #52, opened 2021-08-19
Add support for interfaces, part 1: the simplest cases

benjaminjkraft:
## Summary:
In this commit I begin the journey to add the long-awaited support for
interfaces (part of #8).  Well, it's not the beginning: I already had
some half-written broken code around.  But it's the first fully
functional support, and especially, the first *tested* support; it's
probably best to review the nontrivially-changed code as if it were new.

Conceptually, the code so far is pretty simple: we generate an interface
type, and the implementations.  (That code is in fact mostly unchanged.)
The complexity comes in because encoding/json doesn't know how to
unmarshal that.  So we have to add an UnmarshalJSON method, which
actually has to be on the types with interface-type fields, that knows
how.  I factored it into two methods, such that that UnmarshalJSON
method is just glue, and then there's a separate function, corresponding
to each interface-type, that actually does all the work.  (If only one
could just write it as an actual method!)  The method uses the same
trick suggested to me by a few others in another context to deserialize
all but one field, then handle that field specially, which is discussed
in the code.

This still has some limitations, which will be lifted in future commits:
- it doesn't allow for list-of-interface fields
- it requires that you manually ask for `__typename`
- it doesn't support fragments, i.e. you can only query for interface
  fields, not concrete-type-specific ones
But it works, even in integration tests, which is progress!

As a part of this, I added a proper config option for the "allow broken
features" flag, since I need to be able to set it from the integration
tests which are in a separate package (and actually shell out via `go
generate`).  I also renamed what was to be the first case
(InterfaceNoFragments), and replaced it with a further-simplified
version (avoiding list-of-interface fields.

[1] https://github.com/benjaminjkraft/notes/blob/master/go-json-interfaces.md

## Test plan:
make tesc


generate/testdata/snapshots/TestGenerate-UnionNoFragments.graphql-UnionNoFragments.graphql.go
```diff
@@ -21,9 +22,35 @@ type UnionNoFragmentsQueryRandomLeafLeafContent interface {
...
 }
 
-func (v UnionNoFragmentsQueryRandomLeafArticle) implementsGraphQLInterfaceUnionNoFragmentsQueryRandomLeafLeafContent() {
+func (v *UnionNoFragmentsQueryRandomLeafArticle) implementsGraphQLInterfaceUnionNoFragmentsQueryRandomLeafLeafContent() {
 }
-func (v UnionNoFragmentsQueryRandomLeafVideo) implementsGraphQLInterfaceUnionNoFragmentsQueryRandomLeafLeafContent() {
+func (v *UnionNoFragmentsQueryRandomLeafVideo) implementsGraphQLInterfaceUnionNoFragmentsQueryRandomLeafLeafContent() {
+}
+
+func __unmarshalUnionNoFragmentsQueryRandomLeafLeafContent(v *UnionNoFragmentsQueryRandomLeafLeafContent, m json.RawMessage) error {
```
dnerdy, line 30: Is there an advantage of using `v` as an out argument instead of returning `(InterfaceNestingRootTopicChildrenArticleParentTopicChildrenContent, error)`?

Then the code below would be:
```
err, v.RandomLeaf = __unmarshalUnionNoFragmentsQueryRandomLeafLeafContent(firstPass.RandomLeaf)
```

generate/unmarshal.go.tmpl
```diff
@@ -1,40 +1,40 @@
...
+    {{/* We want to specially handle the abstract fields (.AbstractFields),
+         but unmarshal everything else normally.  To handle abstract fields,
+         first we unmarshal them into a json.RawMessage, and then handle those
+         further, below.  For the rest, we just want to call json.Unmarshal.
+         But if we do that naively on a value of type `.Type`, it will call
+         this function again, and recurse infinitely.  So we make a wrapper
+         type -- with a different name, thus different methods, but the same
+         fields, and unmarshal into that.  For more on why this is so
+         difficult, see
+         https://github.com/benjaminjkraft/notes/blob/master/go-json-interfaces.md
```
dnerdy, line 14: Thanks for including this reference! The unmarshaling is definitely tricky, but it makes sense.

internal/integration/integration_test.go
```diff
@@ -56,6 +56,49 @@ func TestVariables(t *testing.T) {
...
+	defer server.Close()
+	client := graphql.NewClient(server.URL, http.DefaultClient)
+
+	resp, err := queryWithInterfaceNoFragments(ctx, client, "1")
+	require.NoError(t, err)
+
+	assert.Equal(t, "1", resp.Me.Id)
+	assert.Equal(t, "Yours Truly", resp.Me.Name)
+
+	user, ok := resp.Being.(*queryWithInterfaceNoFragmentsBeingUser)
```
dnerdy, line 77: Is there a plan to add helpers so callers don't need to look up the generated type for the type switch? e.g.
```
user, ok := resp.BeingUser()  // or UserBeing()
```

generate/generate_test.go
```diff
@@ -73,9 +71,9 @@ func TestGenerate(t *testing.T) {
 			t.Run("Build", func(t *testing.T) {
 				if testing.Short() {
 					t.Skip("skipping build due to -short")
-				} else if sourceFilename == "InterfaceNesting.graphql" ||
-					sourceFilename == "InterfaceNoFragments.graphql" ||
-					sourceFilename == "Omitempty.graphql" {
+				} else if sourceFilename == "InterfaceNesting.graphql" || // #8
+					sourceFilename == "InterfaceListField.graphql" || // #8
+					sourceFilename == "Omitempty.graphql" { // #43
```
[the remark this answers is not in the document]
benjaminjkraft: Ah, will do!

generate/traverse.go
```diff
@@ -256,10 +256,23 @@ func (g *generator) convertDefinition(
...
+		// We need to request __typename so we know which concrete type to use.
+		hasTypename := false
+		for _, selection := range selectionSet {
+			field, ok := selection.(*ast.Field)
+			if ok && field.Name == "__typename" {
+				hasTypename = true
+			}
+		}
+		if !hasTypename {
+			// TODO(benkraft): Instead, modify the query to add __typename.
```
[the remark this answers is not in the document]
benjaminjkraft: Yeah, this was a way to break up the implementation work, not my idea of how it should look!

generate/testdata/snapshots/TestGenerate-UnionNoFragments.graphql-UnionNoFragments.graphql.go
benjaminjkraft: Ah, interesting.  I was doing this to match json.Unmarshal, and I do think it's marginally more efficient this way, although probably not to an extent that matters.

But I think actually the biggest factor is what will make the most sense when list-of-interface fields make the caller more complicated (#54).  I think actually the out-argument makes things simpler; but maybe you can see what you think when you get there.

internal/integration/integration_test.go
benjaminjkraft: Yep!  There's a TODO somewhere, which is now fixed in #57.

benjaminjkraft:
(@dnerdy just so you know, you never approved this one!  no worries if you want to look again or whatever, just in case you missed it.)

dnerdy (approved):
Whoops. Thanks for the ping!

PR #53, opened 2021-08-20
Correct string zero value in design doc

dnerdy:
The design doc talks about the zero value for the `string` and `*string` types and then references `0` and `null`. This PR changes these values to `""` and `nil`.


benjaminjkraft (approved):
Thanks!

Issue #16, opened 2021-04-22
Make it possible to ask for a *[]T


benjaminjkraft:
If you put `pointer: true` on a list-field, you get `[]*T` (or `[][]*T`, or...) which is probably what you wanted, but not the only option.  If anyone finds a reason to want `*[]T`, or for that matter `*[][]*[]T`, we can figure out a syntax for it; it's easy enough to do technically.

PR #54, opened 2021-08-20
Add support for interfaces, part 2: list-of-interface

benjaminjkraft:
## Summary:
In this commit I remove one of the limitations of our support for
interfaces, from #52, by adding support for list-of-interface fields.
This was surprisingly complex!  The issue is that, as before, it's the
containing type that has to do all the glue work -- and it's that glue
work that is complicated by list-of-interface fields.

All in all, it's not that much new code, and by far the hard part is
just 20 lines in the UnmarshalJSON template (which come with almost
twice as many lines of comments to explain them).  It may be easiest to
start by reading some of the generated code, and then read the template.

I also added support for such fields with `pointer: true` specified,
such that the type is `[][]...[]*MyInterface`, although I don't know why
you would want that.  This does *not* allow e.g. `*[]*[][]*MyInterface`;
that would require a way to specify it (see #16) but also add some extra
complexity (as we'd have to actually walk the type-unwrap chain
properly, instead of just counting the number of slices and whether
there's a pointer).

## Test plan:
make check


generate/testdata/queries/schema.graphql
```diff
@@ -111,6 +111,7 @@ type Query {
   getJunk: Junk
   getComplexJunk: ComplexJunk
   listOfListsOfLists: [[[String!]!]!]!
+  listOfListsOfListsOfContent: [[[Content!]!]!]!
```
dnerdy, line 114: `listOfListsOfListsOfContent: [[[Content]!]!]!` is a type where one might want to use a pointer (though types like this usually seem to usually be accidental). Pointers to slices shouldn't be necessary, so it seems correct to only support pointers on the most-inner type.

generate/testdata/snapshots/TestGenerate-InterfaceListOfListsOfListsField.graphql-InterfaceListOfListsOfListsField.graphql.go
```diff
@@ -0,0 +1,254 @@
...
+		*v = new(InterfaceListOfListOfListsFieldListOfListsOfListsOfContentTopic)
+		return json.Unmarshal(m, *v)
+	default:
+		return fmt.Errorf(
+			`Unexpected concrete type for InterfaceListOfListOfListsFieldListOfListsOfListsOfContent: "%v"`, tn.TypeName)
+	}
+}
+
+// InterfaceListOfListOfListsFieldListOfListsOfListsOfContentArticle includes the requested fields of the GraphQL type Article.
+type InterfaceListOfListOfListsFieldListOfListsOfListsOfContentArticle struct {
```
dnerdy, line 58: Whew. Hopefully no one would actually need to use a type named like this anywhere.

generate/types.go
```diff
@@ -22,6 +22,17 @@ type goType interface {
 	// Reference returns the Go name of this type, e.g. []*MyStruct, and may be
 	// used to refer to it in Go code.
 	Reference() string
+
+	// Remove slice/pointer wrappers, and return the underlying (named (or
+	// builtin)) type.  For example, given []*MyStruct, return MyStruct.
+	Unwrap() goType
+
+	// Count the number of times Unwrap() will unwrap a slice type.  For
+	// example, given []*[]**[]MyStruct, return 3.
```
dnerdy, line 31: nit: this type isn't supported in the current design; it seems like it would be nice to use a valid example (even though the method could in principle support this type as well).

generate/unmarshal.go.tmpl
```diff
@@ -31,13 +31,61 @@ func (v *{{.GoName}}) UnmarshalJSON(b []byte) error {
...
-    {{range .AbstractFields -}}
-    err = __unmarshal{{.GoType.Reference}}(
-        &v.{{.GoName}}, firstPass.{{.GoName}})
-    if err != nil {
-        return err
+    {{/* Now, for each field, call out to the unmarshal-helper.
+         This gets a little complicated because we may have a slice field.
+         So what we do is basically, for each field:
+
+            target := &v.MyField      // *[][]...[]MyType
```
dnerdy, line 38: Same nit here about this not being a supported type.

internal/integration/integration_test.go
```diff
@@ -99,6 +99,36 @@ func TestInterfaceNoFragments(t *testing.T) {
 	assert.Nil(t, resp.Being)
 }
 
+func TestInterfaceListField(t *testing.T) {
```
dnerdy, line 102: Could you also add an integration test that uses `@genqlient(pointer: true)`? I think such a test is the only place the generated code would be exercises (and where we'd suss out any issues). Is that right?

dnerdy (approved):
This looks good to me. I mostly reviewed the generated code and not the template. As mentioned inline, only supporting a pointer for the inner-most type seems right. Also, it would be nice to have an integration test that exercises pointer handling.

generate/testdata/queries/schema.graphql
benjaminjkraft: I think there are valid uses for required-list-of-optional-item, if you want the input in correspondence to the output.  But I agree I can only see a reason to support it at the innermost point.

internal/integration/integration_test.go
benjaminjkraft: My theory was that any issues there would show up at build-time.  But it's probably better not to be lazy :-) .

generate/unmarshal.go.tmpl
benjaminjkraft: This is correct as-is -- `target` has type `*[][]...[]MyType` (or `*[][]...[]*MyType`, because it's a pointer to the field-type, again because that simplifies the templating at the cost of making the generated code look a bit strange.

internal/integration/integration_test.go
benjaminjkraft: LOL, you were write, the code *is* actually broken!  Thanks for the prod :-) .

PR #55, opened 2021-08-20
Use yaml.UnmarshalStrict

benjaminjkraft:
## Summary:
As Mark pointed out in a review where I realized that the *example* had
a bogus config key (which I hadn't noticed earlier because it was just
using the default value), we should probably do a strict-unmarshal;
there's no reason you should have random extra keys in your config and
if you do it's probably a mistake.  (Or maybe you're running a too-old
version of genqlient for your codebase, which you probably also want to
know.)  Now we do.

## Test plan:
- `make check`
- in webapp, `go mod edit -replace github.com/Khan/genqlient=../genqlient`
  then `make genqlient` produces no diffs (except go.mod/go.sum).

### DIFF
--- a/generate/config.go
+++ b/generate/config.go
@@ -120,7 +120,7 @@ func ReadAndValidateConfig(filename string) (*Config, error) {
 			return nil, errorf(nil, "unreadable config file %v: %v", filename, err)
 		}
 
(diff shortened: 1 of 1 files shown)


dnerdy (approved):
Lgtm! 🎉

PR #56, opened 2021-08-23
Automatically add __typename to fields of interface type

benjaminjkraft:
## Summary:
In #52, I added support for interface types, but with the simplifying
restriction (among others) that the user must request the field
`__typename`.  In this commit, I remove this restriction.

The basic idea is simple: we preprocess the query to add `__typename`.
The implementation isn't much more complicated!  Although it required
some new wiring in a few places.

## Test plan:
make tesc


dnerdy (approved):
This looks great! It's nice to have the genqlient automatically request `__typename`.

One very minor issue I can foresee is if an object already has a field named `typename`. Then the names would conflict. Perhaps this means the `__typename` field will need to be made configurable at some point.

benjaminjkraft:
Ah, good point.  I'll just make a separate issue for that; there are some other potential conflicts.

PR #57, opened 2021-08-23
Add support for interfaces, part 4: getter methods

benjaminjkraft:
## Summary:
Right now, if you make a query like `{ myInterface { field } }`, you
have to type-switch on all the possible implementations of `myInterface`
to get at `field`.  Now, we generate getter-methods (e.g. `GetField`),
to make that access easier.  Of course this only applies to shared
fields (which for now are the only ones, but once we support fragments
will no longer be).

This also includes a small change to the way we generate type-names for
interfaces: we no longer include the name of the concrete type in the
interface we propagate forward, so we generate
`MyInterfaceMyFieldMyType`, not `MyInterfaceMyImplMyFieldMyType`, in the
case where you have an interface `MyInterface` implemented by `MyImpl`
(and maybe other types) with field `myField: MyType`.  This is necessary
so the getter method returns a well-defined type, and also probably
convenient for calling code.  It will have to get a little bit more
complicated once we support fragments, where you could have two
implementing types with identically-named fields of different types, but
I think it'll be easiest to figure out how to deal with that when
implementing fragments.

While I was in the area, I added to the interface doc-comment a list of
the implementations.  (In GraphQL, we're guaranteed to know them all
assuming our schema is up to date.)

## Test plan:
make check


internal/integration/integration_test.go
```diff
@@ -116,11 +137,25 @@ func TestInterfaceListField(t *testing.T) {
 
 	require.Len(t, resp.Beings, 3)
 
+	// We should get the following three beings:
```
benjaminjkraft, line 140: note to self: post-rebase, need to apply these changes to `TestInterfaceListPointerField`.

generate/convert.go
```diff
@@ -264,19 +265,52 @@ func (g *generator) convertDefinition(
...
 		}
 		g.typeMap[name] = goType
 
+		// TODO(benkraft): This sorta-duplicates what we'll do in each
+		// implementation when it traverses the fields.  But they'll differ
+		// more once we support fragments; at that point we should figure out
+		// how to refactor.
+		for _, selection := range selectionSet {
+			field, ok := selection.(*ast.Field)
+			if !ok { // fragment/interface, not a shared field
```
dnerdy, line 279: As we discovered with the validator edge case, a union doesn't have shared fields, but it may be possible to treat the fields as shared if all the union members implement the same interface corresponding to a fragment. Anyway, this will be tackled in the upcoming fragment changes (if we want to tackle it at all).

dnerdy (approved):
Looks great! The detailed docstrings are a nice touch.

PR #66, opened 2021-08-26
Fix type-naming in the presence of interfaces, and refactor it a lot

benjaminjkraft:
## Summary:
When adding support for interfaces, I did not do the type-names as I
intended: they came out to be `MyFieldMyType`, not
`MyInterfaceMyFieldMyType`, which is inconsistent, but not strictly
wrong.  But once supporting fragments, this is also now incorrect.
(Exactly why is described in the comments inline.)  In this commit, in
any case, I fix it.

To do that, I finally did the last of the refactors I've been hoping to
do but unable to successfully implement, which is to make the type-name
and type-name-prefix management clearer.  In the past it was kind of
spread out, and each caller would have to pass the right name into
`convertDefinition`, which go quite unwieldy.  Now, the case that really
wanted that -- the operation toplevel -- just does it own thing; and the
main name-generation code  is factored out into a separate file with
tests, and with a long comment that goes into all the details of the
algorithm that the design-doc didn't cover.  (I even had some fun using
a linked list to implement the prefix-stack!)

This allowed me to fix the above bug fairly easily -- actually the fix
was pretty much automatic once I understood how to organize things.
There is one change which is that if your query name is unexported, we
no longer do the same with the input-type names; it's unclear to me if
anyone will actually care about this behavior (Khan always makes the
queries exported) but if they did it was very inconsistent (only at the
query toplevel, and only for input-objects, not enums), so we can
reimplement it properly if that comes up.  As a bonus fix, we now better
handle the case where your type-names are lowercase, which is legal if
nonstandard GraphQL.

## Test plan:
make tesc

### DIFF
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -345,6 +345,8 @@ func (g *generator) convertSelectionSet(
 		case *ast.FragmentSpread:
 			return nil, errorf(selection.Position, "not implemented: %T", selection)
 		case *ast.InlineFragment:
+			// (Note this will return nil, nil if the fragment doesn't apply to
+			// this type.)
 			fragmentFields, err := g.convertInlineFragment(
 				namePrefix, selection, containingTypedef, queryOptions)
 			if err != nil {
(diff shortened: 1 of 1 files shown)


dnerdy (commented):
@benjaminjkraft I'm not sure where the changes from this PR ended up. I see some changes were merged into https://github.com/Khan/genqlient/compare/benkraft.fragments-1, but I don't see the bulk of the changes from this PR on that branch.

(There are some minor inline comments on the original diff below.)

benjaminjkraft:
Oh my, let me make a new PR.  None of this should be merged, but I think when I messed up a previous push I didn't realize everything it had broken.

PR #68, opened 2021-08-27
Return clearer errors when __typename is missing

benjaminjkraft:
## Summary:
In practice, at Khan at least, this is easy to mess up when writing
mocks, because you write the mock by looking at the query, and the query
doesn't say it's asking for `__typename` (because genqlient
automatically adds that).  A sufficiently-smart mocking library might be
able to fix that, or detect it at least, but in any case, we can give a
clearer error.

I also removed an unrelated TODO that was done.

Issue: https://khanacademy.slack.com/archives/C01120CNCS0/p1630019788014000

## Test plan:
make check


dnerdy (approved):
Lgtm!

PR #70, opened 2021-08-28
If requested, validate binding-types get the right fields

benjaminjkraft:
## Summary:
One sharp edge of the new `bindings` setting (when used for composite
types) is this: the (presumably struct) type to which you're binding
may expect to have particular fields, but it's GraphQL so you could have
requested some other set of fields.  Now, if you ask us, we check.

Specifically, I've added a new setting under the `bindings` items, which
says: everywhere we query this must select these fields.  (Or use its
own inline `# @genqlient(bind: ...)`.)  It must select exactly those
fields, in order, no more, no less.  This was fairly easy to implement;
actually comparing the selections was surprisingly much code but it's
all pretty straightforward.

## Test plan:
make check

### DIFF
--- a/generate/binding_validation.go
+++ b/generate/binding_validation.go
@@ -1,0 +1,111 @@
+package generate
+
+// This file is responsible for doing the validation for type-bindings, if they
+// are so configured (see TypeBinding).
+
+import (
+	"fmt"
+
+	"github.com/vektah/gqlparser/v2/ast"
+	"github.com/vektah/gqlparser/v2/parser"
+)
+
+// selectionsMatch recursively compares the two selection-sets, and returns an
+// error if they differ.
+//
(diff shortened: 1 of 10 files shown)


dnerdy (approved):
The implementation looks straight forward. 👍

I was wondering if json.Decoder with DisallowUnknownFields set could work for this. The validation wouldn't be as strong, but maybe it would be good enough? (Also, the expectation for UnmarshalJSON would be that the mention returns an error if the value is somehow incompatible.)

benjaminjkraft:
I think it's better for us to validate, because we can do it at compile-time.  DisallowUnknownFields might be a good idea, although I worry that some GraphQL server will be badly-behaved and it will do more harm than good.

PR #71, opened 2021-08-28
Fix type-naming in the presence of interfaces, and refactor it a lot

benjaminjkraft:
## Summary:
When adding support for interfaces, I did not do the type-names as I
intended: they came out to be `MyFieldMyType`, not
`MyInterfaceMyFieldMyType`, which is inconsistent, but not strictly
wrong.  But once supporting fragments, this is also now incorrect.
(Exactly why is described in the comments inline.)  In this commit, in
any case, I fix it.

To do that, I finally did the last of the refactors I've been hoping to
do but unable to successfully implement, which is to make the type-name
and type-name-prefix management clearer.  In the past it was kind of
spread out, and each caller would have to pass the right name into
`convertDefinition`, which go quite unwieldy.  Now, the case that really
wanted that -- the operation toplevel -- just does it own thing; and the
main name-generation code  is factored out into a separate file with
tests, and with a long comment that goes into all the details of the
algorithm that the design-doc didn't cover.  (I even had some fun using
a linked list to implement the prefix-stack!)

This allowed me to fix the above bug fairly easily -- actually the fix
was pretty much automatic once I understood how to organize things.
There is one change which is that if your query name is unexported, we
no longer do the same with the input-type names; it's unclear to me if
anyone will actually care about this behavior (Khan always makes the
queries exported) but if they did it was very inconsistent (only at the
query toplevel, and only for input-objects, not enums), so we can
reimplement it properly if that comes up.  As a bonus fix, we now better
handle the case where your type-names are lowercase, which is legal if
nonstandard GraphQL.

## Test plan:
make tesc

### DIFF
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -248,17 +248,6 @@ func (g *generator) convertDefinition(
 		g.typeMap[name] = goType
 
 		for i, implDef := range implementationTypes {
-			// Note for shared fields we propagate forward the interface's
-			// name-prefix: that is, the implementations will have fields with
-			// types like
-			//	MyInterfaceMyFieldMyType
-			// not
-			//	MyInterfaceMyImplMyFieldMyType
-			//             ^^^^^^
-			// In particular, this means that the Go type of MyField will be
-			// the same across all the implementations; this is important so
-			// that we can write a method GetMyField() that returns it!
-			// STOPSHIP: update all this nonsense
 			// TODO(benkraft): In principle we should skip generating a Go
 			// field for __typename each of these impl-defs if you didn't
 			// request it (and it was automatically added by
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -11,7 +11,6 @@ package generate
 
 import (
 	"fmt"
-	"strings"
 
 	"github.com/vektah/gqlparser/v2/ast"
 )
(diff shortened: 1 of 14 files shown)


benjaminjkraft:
This is a fixed version of #66.

benjaminjkraft:
(Plus, now, a rebase!)

generate/names.go
```diff
@@ -0,0 +1,157 @@
+package generate
+
+// This file generates the names for genqlient's generated types.  This is
+// somewhat tricky because the names need to be unique, stable, and, to the
+// extent possible, human-readable and -writable.  See DESIGN.md for an
+// overview of the considerations; in short, we need long names.
```
dnerdy, line 6: I said it in #66, but I'll say it again. Amazing documentation! Thanks.

generate/names.go
```diff
@@ -0,0 +1,157 @@
...
+	"github.com/vektah/gqlparser/v2/ast"
+)
+
+// Yes, a linked list!  We could use a stack -- it would probably be marginally
+// more efficient -- but then the caller would have to know more about how to
+// manage it safely.  Using a list, and treating it as immutable, makes it
+// easy.
+type prefixList struct {
+	last string // the list goes back-to-front, so this is the *last* prefix
+	rest *prefixList
```
dnerdy, line 108: nit: perhaps `previous` instead of `rest`? Or maybe name `last` either `prefix` or `name` or `element` and say this is a forward linked list but with the elements in the opposite order?

generate/names.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func joinPrefixList(prefix *prefixList) string {
+	var reversed []string
+	for ; prefix != nil; prefix = prefix.rest {
+		reversed = append(reversed, prefix.last)
+	}
+	l := len(reversed)
+	for i := 0; i < l/2; i++ {
+		reversed[i], reversed[l-1-i] = reversed[l-1-i], reversed[i]
```
dnerdy, line 118: Heh, nice.

generate/convert.go
```diff
@@ -51,20 +50,26 @@ func (g *generator) convertOperation(
...
-		// Query/Mutation is unlikely to be of value.
-		// TODO(benkraft): This is a bit awkward/fragile.
-		structType.Description =
-			fmt.Sprintf("%v is returned by %v on success.", name, operation.Name)
-		structType.Incomplete = false
+	// Instead of calling out to convertType/convertDefinition, we do our own
+	// thing, because we want to do a few things differently, and because we
+	// know we have an object type, so we can include only that case.
+	fields, err := g.convertSelectionSet(
+		&prefixList{last: operation.Name}, operation.SelectionSet, baseType, queryOptions)
```
dnerdy, line 57: It seems like a prefixList constructor could be useful so the caller doesn't need to know anything about the internal structure of the list.

dnerdy (approved):
This is indeed much easier to understand.

generate/names.go
benjaminjkraft: Ah, good ideas.

generate/names.go
benjaminjkraft: I wish I did not have to know that this is the standard way to reverse a list in Go!

generate/convert.go
benjaminjkraft: ah, sure.

PR #74, opened 2021-09-02
Add IDEs to gitignore and add default config file name

### LINKED ISSUE #73: Don't default to no config, and/or document it more clearly
It's arguably confusing that running with no arguments uses no config filename, rather than guessing the filename.  Ideally it should either guess, or demand you specify; but certainly it should tell you in the README what it wants you to do -- the documentation under "Example" is incomplete.  (It will work with the literal query it's using, but not with one that needs a config.)

As a workaround, of course, you can explicitly specify the config, the problem is just that the error you get may not make it obvious that we didn't even look at your config.

StevenACoffman:
Fixes #73

### DIFF
--- a/generate/main.go
+++ b/generate/main.go
@@ -49,7 +49,7 @@ func Main() {
 	case 2:
 		err = readConfigGenerateAndWrite(os.Args[1])
 	case 1:
-		err = readConfigGenerateAndWrite("")
+		err = readConfigGenerateAndWrite("genqlient.yaml")
 	default:
 		argv0 := os.Args[0]
 		if strings.Contains(argv0, string(filepath.Separator)+"go-build") {
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,53 @@
 generate/testdata/tmp
+
+## Go standard gitignore
+# Binaries for programs and plugins
+*.exe
+*.exe~
+*.dll
+*.so
+*.dylib
+
+# Test binary, built with `go test -c`
+*.test
+
+# Output of the go coverage tool, specifically when used with LiteIDE
+*.out
+
+# Dependency directories (remove the comment below to include it)
+# vendor/
+
+## VSCode standard gitignore
+.vscode/*
+!.vscode/settings.json
+!.vscode/tasks.json
+!.vscode/launch.json
+!.vscode/extensions.json
+*.code-workspace
+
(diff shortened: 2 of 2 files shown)


benjaminjkraft (approved):
Thanks!!

generate/main.go
```diff
@@ -49,7 +49,7 @@ func Main() {
 	case 2:
 		err = readConfigGenerateAndWrite(os.Args[1])
 	case 1:
-		err = readConfigGenerateAndWrite("")
+		err = readConfigGenerateAndWrite("genqlient.yaml")
```
dnerdy, line 52: I think this means users must now have a genqlient.yaml file to use genqlient, and [checking for an empty string in ReadAndValidateConfig](https://github.com/Khan/genqlient/blob/6fdb170b99ebb7129675e632baa74dc14d26851a/generate/config.go#L164) is no longer necessary.

Thinking about how to document this for new users, it seems like retaining the ability to run the tool without a config (i.e. falling back to the default config) could still be a good experience. The README could be updated with information about how to override the defaults by providing a config, but users wouldn't need to create or copy a config file to get started.

What do you think about that approach?

benjaminjkraft: One thing we could do is say if you don't specify a config file, then we check `genqlient.yaml` exists, and use it if it does, but proceed if not.  That solves the confusing behavior Steve hit, although maybe it creates new confusing behavior if you misspell the default filename or something.  We could have the examples/documentation encourage you to specify.

Or we could say we log a warning but proceed with default config if no arguments are given.  Or have the filename `-` mean "use the default config" (or something).  I'm pretty flexible here, although I think ideally we want to settle on the behavior sooner rather than later.

dnerdy: What if we require the file to exist but make it easy to create an initial config? We could do this by introducing subcommands, e.g.
```
genqlient init
genqlient generate [CONFIG]
```
That would make it easy for users to get up and running (and the default config could provide pointers into the documentation) while also avoiding any potential confusion.

benjaminjkraft: Great idea!  Steve, why don't you land this, then make a new issue for Mark's suggestion and I can take a look soon.  (Unless you want to go ahead and implement it, of course, I won't mind!)

StevenACoffman: Done!

PR #75, opened 2021-09-03
Add support for concrete-typed named fragments

benjaminjkraft:
## Summary:
In previous commits I added support to genqlient for interfaces and
inline fragments.  This means the only query structures that remain are
named fragments and their spreads, e.g.
```
fragment MyFragment on MyType { myField }
query MyQuery { getMyType { ...MyFragment } }
```
Other than mere completionism, these are potentially useful for code
sharing: you can spread the same fragment multiple places; and then
genqlient can notice that and generate the same type for each.  (They
can even be shared between different queries in the same package.)

In this commit I add support for named fragments of concrete
(object/struct, not interface) type, spread into either concrete or
abstract scope.  For genqlient's purposes, these are a new "root"
type-name, just like each operation, and are then embedded into the
appropriate struct.  (Using embeds allows their fields to be referenced
as fields of the containing type, if convenient.  Further design
considerations are discussed in DESIGN.md.)

This requires new code in two main places (plus miscellaneous glue),
both nontrivial but neither particularly complex:
- We need to actually traverse both structures and generate the types
  (in `convert.go`).
- We need to decide which fragments from this package to send to the
  server, both for good hyigene and because GraphQL requires we send
  only ones this query uses (in `generate.go`).
- We need a little new wiring for options -- because fragments can be
  shared between queries they get their own toplevel options, rather
  than inheriting the query's options.

Finally, this required slightly subtler changes to how we do
unmarshaling (in `types.go` and `unmarshal.go.tmpl`).  Basically,
because embedded fields' methods, including `UnmarshalJSON`, get
promoted to the parent type, and because the JSON library ignores their
fields when shadowed by those of the parent type, we need a little bit
of special logic in each such parent type to do its own unmarshal and
then delegate to each embed.  This is similar (and much simpler) to
what we did for interfaces, although it required some changes to the
"method-hiding" trick (used for both).  It's only really necessary in
certain specific cases (namely when an embedded type has an
`UnmarshalJSON` method or a field with the same name as the embedder),
but it's easier to just generate it always.  This is all described in
more detail inline.

This does not support fragments of abstract type, which have their own
complexities.  I'll address those, which are now the only remaining
piece of #8, in a future commit.

## Test plan:
make check


generate/types.go
```diff
@@ -144,28 +152,47 @@ func (typ *goStructType) WriteDefinition(w io.Writer, g *generator) error {
...
-	// Specifically, in order to unmarshal interface values, we need to add an
-	// UnmarshalJSON method to each type which has an interface-typed *field*
-	// (not the interface type itself -- we can't add methods to that).
-	// But we put most of the logic in a per-interface-type helper function,
-	// written along with the interface type; the UnmarshalJSON method is just
-	// the boilerplate.
-	if len(typ.AbstractFields()) == 0 {
+	// For embedded fields (from fragments), mostly the JSON library would just
+	// do what we want, but there are two problems.  First, if the embedded
+	// ype has its own UnmarshalJSON, naively that would be promoted to
```
dnerdy, line 186: typo: "ype" => type

generate/convert.go
```diff
@@ -354,7 +359,9 @@ func (g *generator) convertSelectionSet(
 		// GraphQL (and, effectively, JSON) requires that all fields with the
 		// same alias (JSON-name) must be the same (i.e. refer to the same
 		// field), so that's how we deduplicate.
-		if fieldNames[field.JSONName] {
+		// It's fine to have duplicate embeds (i.e. via named fragments), even
+		// ones with complicated overlaps, since they are separate types to us.
```
dnerdy, line 363: It might be worth explicitly calling out that field.JSONName == "" indicates an embed.

dnerdy (approved):
This came together really nicely! It seems like a surprisingly small amount of code for what it does.

csilvers:
I want to keep track of when this lands, and I don't know how to do that without joining the thread, so I'm making this contentless comment!

generate/convert.go
```diff
@@ -444,6 +451,73 @@ func (g *generator) convertInlineFragment(
 		containingTypedef, queryOptions)
 }
 
+// convertFragmentSpread converts a single GraphQL fragment-spread
+// (`...MyFragment`) into a Go struct-field.  It assumes that
+// convertNamedFragment has already been called on the fragment-definition.  If
```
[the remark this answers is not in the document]
benjaminjkraft: Ah, yes, I changed this to avoid needing to topo-sort the fragments!

generate/convert.go
benjaminjkraft: This will actually be further clarified in the next commit, so I'm gonna leave it to avoid conflicts.

benjaminjkraft:
@csilvers I believe you can subscribe with the button in the sidebar -- this one (except it will be in the opposite toggle-state)
GitHub "unsubscribe from notifications" button

csilvers:
Which sidebar is that?  Note that I only want to subscribe to this PR, not to the entire repo.

benjaminjkraft:
On the PR!  Yours will look more like:
Screenshot 2021-09-09 at 18-25-06 Khan webapp The Khan Academy web app

PR #77, opened 2021-09-03
Clean up, test, and document ContextType and ClientGetter options

### LINKED ISSUE #5: Clean up client_getter (and document it)
One option is to just say it's a function whose argument is context.  (That way we can handle imports right.)  But ideally it could also be a global var; and what if you want to be able to return an error?

Or we could just flag it out for now, since Khan didn't end up needing it.

benjaminjkraft:
## Summary:
ContextType is in use at Khan as a part of our ka-context system; it
basically just lets you configure the type to pass as the `ctx` argument
to genqlient helpers (or say to omit such an argument).  ClientGetter I
wrote thinking we might use it; then we didn't (because we have a few
different clients we may use) but it's not much code and may be helpful
to others.  In this commit I clean up, document, and add tests for both
options.

The cleanup is mainly for ClientGetter, which was kind of broken before
because it was a Go snippet but couldn't specify imports.  I was
thinking maybe you want to be able to write `ctx.Something()`, but I
just don't see how to make it work, so I made it a function of context,
which is probably the better idea anyway.

Additionally, I improved the documentation for both, and added tests for
those and several other config options that weren't completely tested.

Fixes #5.

Issue: https://github.com/Khan/genqlient/issues/5

## Test plan:
make check


generate/generate_test.go
```diff
@@ -91,29 +121,103 @@ func TestGenerate(t *testing.T) {
...
+			t.Run("Build", func(t *testing.T) {
+				if testing.Short() {
+					t.Skip("skipping build due to -short")
 				}
 
-				cmd := exec.Command("go", "build", f.Name())
-				cmd.Stdout = os.Stdout
-				cmd.Stderr = os.Stderr
-				err = cmd.Run()
+				fmt.Println(generated)
```
dnerdy, line 216: I suspect you don't want this Println here.

dnerdy (approved):
Looks great!

PR #80, opened 2021-09-08
Create a separate module for our golangci-lint version

### LINKED ISSUE #62: Clean up go.mod
Right now it has a bunch of test-only deps, which is mildly annoying but hopefully not a huge deal, but also golangci-lint, which is bad because you really want that pinned to a specific *minor* version, and now pulling in genqlient might force you to upgrade it.  (Arguably that's golangci-lint's problem, but I think it's just life when it comes to a lint tool; honestly we will have similar problems, see #63.)  We should definitely fix that for golangci-lint, which is very doable since all we do is `go run` it, we don't actually need to import it.  Ideally we also fix it for all the test-only deps (cupaloy, testify, gqlgen), although the latter may be swimming too much upstream of what go.mod wants us to do.  Maybe it's doable at least for gqlgen, which is by far the largest of these in terms of import graph, and is integration-test-only.

benjaminjkraft:
## Summary:
In principle, it's not a problem to have test-only deps in your go.mod,
because they won't end up in your importers' builds (and in newer Go
versions may not even be downloaded.  (Which is why there's no
annotation to do so.)  In practice, that doesn't really work for
golangci-lint, because it doesn't really use semver (reasonably, in that
any updated linter may break lint in your codebase).  And we don't
really need it other than to run the binary at a particular version.

So now, I put it in its own go module.  This requires a bit more
throat-clearing to run it (we can't just `go run`), but it's not so bad
and avoids anyone getting annoyed at us because we upgraded their
golangci-lint for them.

We could do the same for `internal/integration`, which adds quite a lot,
including gqlgen, to our dependency tree, but it's not clear there's a
need.

Fixes #62.

Issue: https://github.com/Khan/genqlient/issues/62

## Test plan:
make check


benjaminjkraft:
(@StevenACoffman, I cc'ed you here in case you know a better way to do this, or know of reasons why I should do it for gqlgen as well.)

dnerdy (approved):
Makes sense to me!

internal/lint/tools.go
```diff
@@ -0,0 +1,14 @@
+// Conventionally this sort of thing would use the "ignore" tag, but
+// `go mod tidy` ignores so-tagged files explicitly, so we use another build
+// tag we never intend to set.
+// +build tools
+//go:build tools
```
StevenACoffman, line 5: So is `go:build` in Go 1.17? I remembered something about "finishing the transition in Go 1.18" (which maybe was just dropping `// +build` support)?

I also recall that some old versions of Go required `// +build tools` to be the *very first line* of the file... but that might have been before `go mod` so maybe not an issue? 🤷‍♂️ I think the current rules (until 1.18) are:

1. start with the prefix +build followed by one or more spaces
2. located at top of file *before* the package declaration
3. have at least one blank line between it and package declaration to prevent it from being treated as package documentation
4. The tags separated by space will be interpreted under the OR logic.
5. Comma separated tags will be interpreted under the AND logic.
6. Each term is an alphanumeric word and if preceded by ! it means it is negated.

internal/lint/go.mod
```diff
@@ -0,0 +1,6 @@
+module github.com/Khan/genqlient/internal/lint
+
+go 1.14
```
StevenACoffman, line 3: 1.14 is *really* old. I kind of think Khan is close to the trailing edge when it comes to Go version support. Most surveys show that 80% are on the current or last version by 6 months after a release. With container based runtimes and dependabot, it's automagical.

StevenACoffman (approved):
This is how I would do it, but are we doing this to support webapp dependencies' updates lagging?

internal/lint/tools.go
benjaminjkraft: Yeah, I think 1.16-1.18 all sync the two, and 1.17 is where `go:build` becomes the source of truth.  Since it's easy, I plan to keep both around until we have other reasons to drop 1.16, which is presumably a while yet.

With all that said, I could probably remove this build tag now; nobody should be trying to build anything in this module!

internal/lint/go.mod
benjaminjkraft: Yeah, I think that's accurate.  But my general feeling is to wait to drop support until there's a nontrivial reason to, which is why I dropped 1.13 recently, but kept 1.14.

benjaminjkraft:
No, Craig actually just updated golangci-lint in webapp when pulling this in, so it was not an issue in practice at Khan.  But I still felt it was a potential problem!

Issue #31, opened 2021-04-22
Add a fancier config/arguments setup


benjaminjkraft:
Could use something like [viper](https://github.com/spf13/viper).  IMO it's much better to have it in a file for most cases, but you could imagine there are times when it's nice to not.

benjaminjkraft:
Ok, I looked into this and I'm not sure something like viper would provide value right now.  If someone finds it useful to pass config on the CLI, we can look into it then; otherwise this is fixed by introducing go-args in #81.

PR #81, opened 2021-09-08
Add a flag --init to write a default config

### LINKED ISSUE #76: Add subcommands to init config file, and otherwise require the config file exists
Follow-up to #73 : Require the file to exist but make it easy to create an initial config? We could do this by introducing subcommands, e.g.:
```
genqlient init
genqlient generate [CONFIG]
```
That would make it easy for users to get up and running (and the default config could provide pointers into the documentation) while also avoiding any potential confusion.

benjaminjkraft:
## Summary:
Steve pointed out (#73) that having genqlient with no arguments silently
use a default config file was a bit confusing, and changed it to use
`genqlient.yaml` by default (#74).  Mark pointed out (#76) that this
makes it a bit less convenient when you're starting from scratch; you
have to go create a config file.  In this commit I add a new init flag
that creates you a config file before using it.

Originally the suggestion was to use subcommands, e.g. we'd have
`genqlient init` and `genqlient generate` and so on.  But I couldn't
think of anything else we might want subcommands for in the future, and
it felt a little silly to make you type `generate` each time.  So
instead, I made it a flag, which has the nice property that you can do
`genqlient --init` and it will generate and then use a config file.  (I
mean, maybe it will immediately crash because you don't have a schema,
but hopefully that's still a useful clue as to what to do next!)  The
implmentation was fairly trivial.

Since we now have a nice way to generate a default config, I removed the
default values for most of the options; I've always felt they were
probably more confusing than helpful.  (And indeed, all the users I know
of (Khan/webapp, and the much smaller project Steve was working on, are
setting those options explicitly.)  This required a slight change to
the syntax to say "don't use context", which is probably also net clearer.

I decided this is also a good time to pull in a proper CLI parser (#31);
see ADR-504 for more on that choice.  This also adds some nice help
messages!

Fixes #76, #31.

Issue: https://github.com/Khan/genqlient/issues/76

## Test plan:
```
go run .
go run . --init
go run . --init example/genqlient.yaml     # refuses to clobber
go run . --init example/newgenqlient.yaml
```


generate/generate_test.go
```diff
@@ -130,6 +132,22 @@ func TestGenerate(t *testing.T) {
 	}
 }
 
+func defaultConfig(t *testing.T) *Config {
+	// Parse the config that `genqlient --init` generates, to make sure that
+	// works.
```
dnerdy, line 137: Nice.

generate/main.go
```diff
@@ -36,25 +38,32 @@ func readConfigGenerateAndWrite(configFilename string) error {
 	return nil
 }
 
+type cliArgs struct {
+	ConfigFilename string `arg:"positional" placeholder:"FILENAME" default:"genqlient.yaml" help:"path to genqlient configuration (default genqlient.yaml)"`
```
dnerdy, line 42: Perhaps `placeholder:"CONFIG"`?

dnerdy (approved):
Sub-commands could be useful for things like exporting operations, but that's already supported in the config (and similar things could be added there as well).

I do like how this turned out. I also like explicit context suppression using "-".

Issue #26, opened 2021-04-22
Revamp documentation


benjaminjkraft:
We need some proper documentation, not just random bits and bobs in the README and godoc.  This can be based in part on what people hit in use at Khan.

Incomplete list: 
- [x] adapt internal Khan doc "Using genqlient in webapp"
- [x] FAQ on "how do I represent/do/handle ..." to guide people to which options they need

PR #84, opened 2021-09-09
Reorganize documentation to make room to grow

benjaminjkraft:
## Summary:
In this commit I reorganize much of our documentation into a new `docs`
directory, where there will hopefully be more room to grow and to
organize things in a user-friendly way.  There's almost no net-new
documentation, although of course it's a great time to review it anyway.

In particular:
- I moved the documentation for the `genqlient.yaml` config file into an
  example file instead of GoDoc (which now just points to the example
  file); I think this will be a lot clearer for casual users.
- I moved the documentation for the `@genqlient` directive out of GoDoc
  and into a GraphQL schema file (since while it's a comment it's all
  real syntax), likewise, and made the `GenqlientDirective` type private
  (since there's now nothing useful to do with it).
- I moved `DESIGN.md` and the logo into `docs/` (just to keep the
  toplevel a bit cleaner), and separated the Contributing section of the
  README into `docs/CONTRIBUTING.md` (which github will automatically
  link on various issue and PR pages).

This leaves it so that:
- README.md is the only documentation at the toplevel (and will become
  just the high-level introduction as I add more user docs to `docs/`)
- GoDoc is only documentation for if you want to call genqlient
  programmatically (which is fairly limited as the API surface is quite
  small: it's now just Main, Generate, and Config, plus a constructor, a
  single method, and a bunch of fields on the latter)

In future commits, I'll add some more new documentation to the `docs`
directory.

Issue: https://github.com/Khan/genqlient/issues/26

## Test plan:
make check (and read the docs)


docs/genqlient.yaml
```diff
@@ -0,0 +1,114 @@
...
+# The package name for the output code; defaults to the directory name of
+# the generated-code file.
+package: mygenerated
+
+# If set, a file at this path (relative to genqlient.yaml) will be generated
+# containing the exact operations that genqlient will send to the server.
+#
+# This is useful for systems which require queries to be explicitly
+# safelisted, especially for cases like queries involving fragments where
+# it may not exactly match the input queries, or for other static analysis.
```
jvoll, line 32: Is safelisting queries a standard GraphQL practice or a KA practice? Will others know why you might safelist queries?

jvoll (approved):
LGTM

docs/genqlient.yaml
benjaminjkraft: Ah good point.  I think its pretty common but I can link to e.g. https://www.apollographql.com/docs/studio/operation-registry/

benjaminjkraft:
For other reviewers: I'm landing this now to avoid conflicts, but reviews later are still welcome!

Issue #39, opened 2021-04-23
Document behavior/options for the 32-bit int issue


benjaminjkraft:
- note that GraphQL ints are 32-bit, servers may reject or misbehave:
  - apollo-server explicitly checks ints are 32 bits
  - gqlgen checks (via ParseInt) that ints fit in whatever type your code uses (default int)
  - graphene documents 53 bits but actually requires 32 (unless you use its BigInt type)
- default behavior is to not check
  - apollo-client doesn't check (although it's JS so it's at most 53 bits)
  - shurcooL effectively makes you pass its `graphql.Int` type, which is `int32`
- instead can set an `int32` in scalars, or a custom type that say warns on conversion

PR #86, opened 2021-09-09
Add a bunch more documentation (and rewrite README)

### LINKED ISSUE #26: Revamp documentation

benjaminjkraft:
## Summary:
In this commit, I add a bunch more documentation of how to use
genqlient, including a getting started guide, and a FAQ that also acts
as a how-to shaped index of common configuration options and extension
points.  Much of this is adapted from the internal doc I wrote at Khan,
or reorganized from what was in the README.  Speaking of which, now that
there are better places for all the details, I rewrote the README to be
a bit more of an overview and index of other documentation.  (For now I
left the "unmaintained" notes, which I'll remove fairly soon once all
our ducks are in a row.)

Fixes #26, #39.

Issue: https://github.com/Khan/genqlient/issues/26
Issue: https://github.com/Khan/genqlient/issues/39

## Test plan:
make check, read the docs

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -91,7 +91,7 @@ Or, you can bind it to any other type, perhaps one with size-checked constructor
 genqlient supports GraphQL field-aliases, and uses them to determine the Go struct field name.  For example, if you do
 ```graphql
 query MyQuery {
-  myGreatName: myField
+  myGreatName: myString
 }
 ```
 and genqlient will generate a Go field `MyGreatName string`.  Note that the alias will always be uppercased, to ensure the field is visible to the Go JSON library.
@@ -102,11 +102,11 @@ There are two ways to handle nullable fields in genqlient.  One way is to use th
 
 ```graphql
 query MyQuery(arg: String) {
-  myField
+  myString
 }
 ```
 
-then genqlient will generate a Go field `MyField string`, and set it to the empty string if the server returns null.  This works even for structs: if an object type in GraphQL is null, genqlient will set the corresponding struct to its zero value.  It can be helpful to request `id` in such cases, since that’s a field that should always be set, or `__typename` which is guaranteed to be set, so you can use its presence to decide whether to look at the other fields.
+then genqlient will generate a Go field `MyString string`, and set it to the empty string if the server returns null.  This works even for structs: if an object type in GraphQL is null, genqlient will set the corresponding struct to its zero value.  It can be helpful to request `id` in such cases, since that’s a field that should always be set, or `__typename` which is guaranteed to be set, so you can use its presence to decide whether to look at the other fields.
 
 For input fields, you often want to tell genqlient to send null to the server if the argument is set to the zero value, similar to the JSON `omitempty` tag.  In this case, you can do:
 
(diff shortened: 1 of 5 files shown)


docs/FAQ.md
```diff
@@ -0,0 +1,236 @@
...
+
+### … a specific name for a field?
+
+genqlient supports GraphQL field-aliases, and uses them to determine the Go struct field name.  For example, if you do
+```graphql
+query MyQuery {
+  myGreatName: myField
+}
+```
+and genqlient will generate a Go field `MyGreatName string`.  Note that the alias will always be uppercased, to ensure the field is visible to the Go JSON library.
```
mahtabsabet, line 97: This is assuming that `myField` was already a string, right?

docs/FAQ.md
```diff
@@ -0,0 +1,236 @@
...
+
+You can also put the `# @genqlient(omitempty: true)` on the first line, which will apply it to all arguments in the query.
+
+If you need to distinguish null from the empty string (or generally from the Go zero value of your type), you can tell genqlient to use a pointer for the field or argument like this:
+```graphql
+query MyQuery(
+  # @genqlient(pointer: true)
+  arg: String,
+) {
+  # @genqlient(pointer: true)
```
mahtabsabet, line 130: Can we also put `# @genqlient(pointer: true)` on the first line to apply it to everything in the query, similar to the previous omitempty example?

docs/FAQ.md
```diff
@@ -0,0 +1,236 @@
...
+      language
+    }
+  }
+}
+```
+
+genqlient will generate the following types:
+
+```go
+type GetBooksFavoriteBook interface {
```
mahtabsabet, line 160: This might be documented elsewhere, but how does genqlient generate the `FavouriteBook` part of the name?

docs/FAQ.md
```diff
@@ -0,0 +1,236 @@
...
+```
+
+### … documentation on the output types?
+
+For any GraphQL types or fields with documentation in the GraphQL schema, genqlient automatically includes that documentation in the generated code's GoDoc.  To add additional information to genqlient entrypoints, you can put comments in the GraphQL source:
+
+```graphql
+# This query gets the current user.
+#
+# @genqlient(...)
```
mahtabsabet, line 191: what is supposed to go inside `@genqlient(...)`?

mahtabsabet (approved):

docs/FAQ.md
benjaminjkraft: Ah, yeah, I'll add some notes to that effect.  Or maybe just call it `myString`!

docs/FAQ.md
benjaminjkraft: Yes!  Will mention.

docs/FAQ.md
benjaminjkraft: Yep it's below!  I can mention it.

docs/FAQ.md
benjaminjkraft: Ah, this is just to show that if you have a `genqlient` directive, you can put it after the documentation.  I'll clarify.

PR #87, opened 2021-09-09
Make README logo work in dark mode

### LINKED ISSUE #17: Image in README is hard to read in GitHub's dark/dimmed mode
Screen Shot 2021-04-21 at 9 43 41 PM
Screen Shot 2021-04-21 at 9 42 36 PM

benjaminjkraft:
## Summary:
Dark mode is, they say, the new thing, and the logo doesn't work super
well with it, because it's black on dark grey.  This is surprisingly
hard to fix.  There are a few options:
- make the logo colored; sorry, but I like black
- add a white border to the logo; this is what Apollo does but I
  think it's pretty ugly in dark mode
- add svg with media queries; this works well but only if the github
  dark mode setting matches the browser(/OS) dark mode setting, since
  that's what the media queries look at
I opted for option 3.  This required converting the text in the SVG to
paths, since it really matters that you have the exact same font.  (I
tested on Android Firefox, which empirically doesn't have the same fonts
I have on desktop.)

Fixes #17.

Issue: https://github.com/Khan/genqlient/issues/17

## Test plan:
tested that it looks good on both mobile firefox in dark mode, and
desktop firefox+chrome in light mode.


dangoor:
Fancy solution! Works well for me. I switched my Mac from dark mode to light mode, and could see the difference with a page reload.

dangoor (approved):

dbraley (approved):
Hm, didn't actually work for me, but I suspect I need to restart my machine for the changes to take effect. I like the solution though, so if Kevin is saying it works for him, it's good by me!

somewhatabstract (approved):
Looks great! Nice!

PR #89, opened 2021-09-11
Clean up example a little

benjaminjkraft:
## Summary:
I realized:
- it's simpler to avoid the usual `cmd` and just put it all in the same
  directory
- we don't need schema.json since GitHub now posts a schema in SDL
  format

## Test plan:
make lint example


dnerdy (approved):
👍

StevenACoffman (approved):
Thanks!

PR #91, opened 2021-09-11
Picking some nits that my IDE complained about

StevenACoffman:

### DIFF
--- a/example/cmd/example/main.go
+++ b/example/cmd/example/main.go
@@ -1,7 +1,7 @@
 package main
 
 import (
-	example "github.com/Khan/genqlient/example"
+	"github.com/Khan/genqlient/example"
 )
 
 func main() {
--- a/generate/names.go
+++ b/generate/names.go
@@ -137,7 +137,7 @@ func joinPrefixList(prefix *prefixList) string {
 // comment), but it's used to construct both the type-names from the input and
 // the next prefix-list.
 func typeNameParts(prefix *prefixList, typeName string) *prefixList {
-	// GraphQL's types are conventionally UpperCamelCase, but it's not required;
+	// GraphQL types are conventionally UpperCamelCase, but it's not required;
 	// our names will look best if they are.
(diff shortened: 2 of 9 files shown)


example/cmd/example/main.go
```diff
@@ -1,7 +1,7 @@
 package main
 
 import (
-	example "github.com/Khan/genqlient/example"
+	"github.com/Khan/genqlient/example"
```
benjaminjkraft, line 4: (This will be removed in #89.)

generate/names.go
```diff
@@ -137,7 +137,7 @@ func joinPrefixList(prefix *prefixList) string {
 // comment), but it's used to construct both the type-names from the input and
 // the next prefix-list.
 func typeNameParts(prefix *prefixList, typeName string) *prefixList {
-	// GraphQL types are conventionally UpperCamelCase, but it's not required;
+	// GraphQL's types are conventionally UpperCamelCase, but it's not required;
```
benjaminjkraft, line 140: I think this is clearer (and correct) as-is?  I think of them as "GraphQL types"; "GraphQL" is the kind of type they are, not the owner of the type.

graphql/util.go
```diff
@@ -6,7 +6,7 @@ package graphql
 // NoUnmarshalJSON is intended for the use of genqlient's generated code only.
 //
 // It is used to prevent a struct type from inheriting its embed's
-// UnmarshalJSON method: given a type
+// UnmarshalJSON method so for a given type:
```
benjaminjkraft, line 9: This doesn't flow as well into the continuation below, I think it's better as-is.  Maybe even better would be:
```suggestion
// UnmarshalJSON method, so if we construct a type:
```
and/or rephrase the subsequent few lines.

benjaminjkraft (approved):
Ok -- a couple of these I don't like and I'm not sure what the IDE is complaining about, but the rest seem fine.

benjaminjkraft (approved):

Issue #90, opened 2021-09-11
nit: TestData snapshots with `.go` extension are not valid Go files


StevenACoffman:
Not a big deal, but some of the files under `generate/testdata/snapshots` have `*.go` extensions but are not valid Go files. Perhaps we could call them `*.snapshot` ?
```
TestGenerateErrors-BindingWithIncorrectSelection.go
TestGenerateErrors-ConflictingSelections.go
TestGenerateErrors-InvalidQuery.go
TestGenerateErrors-InvalidScalar.go
TestGenerateErrors-InvalidSchema.go
TestGenerateErrors-MissingTypeName.go
TestGenerateErrors-NoQuery.go
TestGenerateErrors-UnknownScalar.go
```
In other nits, the `parsing_errors` package contains an underscore, which if we could name it `unparseable` or move it under `parsing` and just call it `errors` it would make that particular linter happy.

benjaminjkraft:
Heh, sure, I can add a `.txt`, although you'll likely want to teach the linter to ignore `testdata` directories as we make no guarantee the generated code will be lint-clean.

PR #92, opened 2021-09-14
Avoid putting incorrect extensions on error snapshot files

benjaminjkraft:
## Summary:
They were accidentally ending up as `.go` or `.graphql`, but are
actually just text.

Issue: https://github.com/Khan/genqlient/issues/90

## Test plan:
make check


StevenACoffman (approved):
Thanks!

Issue #93, opened 2021-09-15
Be less strict about selection-set ordering


benjaminjkraft:
For both the `bindings` option, and the soon-to-be-added `typename` option, we validate that two selection sets are equal.  Right now we demand that they even have the fields in the same order; ideally we'd allow any equivalent selections.  That's surprisingly much work, although it's quite well separated from the rest of the codebase (just modify `selectionsMatch` in `generate/validation.go`).

Issue #60, opened 2021-08-24
Potential field-name conflicts


benjaminjkraft:
The function from GraphQL field-name to Go struct field-name is not injective, so there's the potential that we generate two fields with the same name (which will fail to compile).  The most likely way this would happen is if you request a field `typename` in a selection of interface type, where genqlient will add `__typename`; both translate to `Typename` in Go.  But it's also possible that you request two fields `MyField` and `myField`, or `myField` and `_myField`; by convention field names and aliases are `lowerCamelCase` but it's [not required](https://spec.graphql.org/draft/#Name) and only double-underscore is officially reserved.  It's always possible to work around this by aliasing, of course.

benjaminjkraft:
I think it's going to be too hard to avoid this perfectly, but I have a change in the works to at least error in such cases.

Issue #14, opened 2021-04-22
Improve options-handling for input types


benjaminjkraft:
Right now you can't apply an option, like `pointer: true`, to a field of an input type, because there's nowhere to put it.  So we need to figure out a syntax; maybe just `pointerfields: ["field1", "field2.field3"]` but that will be a bit of an explosion of options.

We also need to decide how to resolve conflicts: this would mean that two references to the same GraphQL input-type are no longer guaranteed to generate the same Go type, which means we have to prefix the names or something.  (This is unlikely to come up within a single query, but more likely across queries.)

As a sort of workaround, you can construct the type inline:
```graphql
query MyQuery(inputField: String!, ...) {
  myField(input: {
    field: $inputField,
    ....
  })
}
```

benjaminjkraft:
Another related problem: you might want to be able to apply `omitempty` to a particular, or all, fields of the input object.  Right now it only applies to arguments.  This will actually require new code to handle!

csilvers:
This is, somewhat surprisingly, turning out to be a non-trivial barrier to genqlient adoption.

I thought of another suggestion, that maybe avoids some of the problems above: allow a user to specify the input fields they're interested in much like (in graphql) they specify the output fields.  The code could be something like this:
```

func QueryMyOperation(...) {
   _ = `# @genqlient(inputFor: "MyOperationName")
   type MyOperationInput {
      email
      # @genqlient(pointer: true)
      validate
   }
   `
   _ = `#genqlient
  query MyOperationName($input: MyOperationInput) {
     user(input: $input) { id }
  `
  genqlient.MyOperationName(...)
```

Then genqlient would validate that the `inputFor` is a subset of the actual graphql type MyOperationInput, and would generate its own type for this, called MyOperationNameMyOperationInput or something.  And that own type would follow the pointer/omitempty/etc directives.  And then you could use that type name as an input to QueryMyOperation, or construct it inside QueryMyOperation, or whatever.

benjaminjkraft:
Oh, that's a neat idea!  I'll take a look at how to make it work.

benjaminjkraft:
As I mention in the above-linked committ, the new `bindings` option is also a sort of workaround to this.

benjaminjkraft:
Ugh, I think this is still worth doing, but the syntax is proving to be a mess.

I like Craig's idea above -- which I would modify to saying
```graphql
fragment InputForMyQuery on MyInput {
  email
  # @genqlient(pointer: true)
[...]

Issue #12, opened 2021-04-22
Allow specifying type-names


benjaminjkraft:
The autogenerated names can be annoying, so you might want to explicitly specify the names.  (With that comes some responsibility: you have to avoid conflicts.  But in practice it's not super hard.)  This would also make #11 less necessary.

PR #94, opened 2021-09-15
Add support for specifying type-names, and conflict-detection

benjaminjkraft:
## Summary:
In this commit I add two related features to genqlient:
conflict-detection to avoid generating two distinct types with the same
name, and an option to specify the type-name genqlient should use for
some type.

The conflict-detection was pretty simple once I realized I had already
written all the code to do it in #70.  There was a bunch of wiring,
since we now need to keep track of the GraphQL type/selection-set that
each type corresponds to, but it was pretty straightforward.  This
allows us to:
- detect and reject if you have really sneaky type-names (there are some
  examples documented in `names.go`)
- more clearly crash if genqlient accidentally generates two conflicting
  types, and
- avoid stack-overflow when handing recursive (input) types (although
  sadly the poor support for options on input types (#14) makes them
  difficult to use in many cases; you really need to be able to set
  `pointer: true`)

And with that all set up, the type-naming was also easy!  (It doesn't
have to get into the core of the type-generator, just plug in where we
choose names.  The desire for conflict detection was the main reason I
hadn't set it up already.)  Note that the existing limitation of #70 that
the fields have to be in exactly the same order remains (and is now
documented as #93); it's not deeply hard to fix but it's surprisingly
much work.

Issue: https://github.com/Khan/genqlient/issues/60
Issue: https://github.com/Khan/genqlient/issues/12

## Test plan:
make check


StevenACoffman (approved):
Love the doc improvements too!

docs/FAQ.md
```diff
@@ -66,6 +66,18 @@ if errors.As(err, &errList) {
...
+
+Just tell genqlient via the `bindings` option in `genqlient.yaml`:
+
+```yaml
+bindings:
+  DateTime:
+    type: time.Time
+```
+
+Make sure the given type has whateevr logic is needed to convert to/from JSON (e.g. `MarshalJSON`/`UnmarshalJSON` or JSON tags).  See the [`genqlient.yaml` documentation](genqlient.yaml) for the full syntax.
```
jvoll, line 79:
```suggestion
Make sure the given type has whatever logic is needed to convert to/from JSON (e.g. `MarshalJSON`/`UnmarshalJSON` or JSON tags).  See the [`genqlient.yaml` documentation](genqlient.yaml) for the full syntax.
```

generate/convert.go
```diff
@@ -15,6 +15,62 @@ import (
...
+			pos, "conflicting definition for %s; this can indicate either "+
+				"a genqlient internal error, a conflict between user-specified "+
+				"type-names, or some very tricksy GraphQL field/type names: "+
+				"expected GraphQL type %s, got %s",
+			goName, typ.GraphQLTypeName(), graphQLName)
+	}
+
+	expectedSelectionSet := typ.SelectionSet()
+	if err := selectionsMatch(pos, selectionSet, expectedSelectionSet); err != nil {
+		fmt.Println(goName, err)
```
jvoll, line 46: Did you intend to both print + return the error?

jvoll (approved):

generate/convert.go
benjaminjkraft: I sure didn't!

Issue #35, opened 2021-04-22
Set up OSS stuff


benjaminjkraft:
- [x] Add to khan.github.io (Khan/khan.github.io#17)
- [x] Contribution guidelines
  - [x] CLA (#96)
  - [x] COC (#96)
  - [x] Contributor-focused docs (#96)
  - [x] "Good first issue"/"help wanted"
- [x] Issue templates (#96)
- [ ] tag a v0.1.0

kevinbarabash:
The CLA form lives at https://www.khanacademy.org/r/cla, but our CLA bot is currently out of commission.

PR #96, opened 2021-09-15
Add open-source boilerplate

benjaminjkraft:
## Summary:
There's a lot these days!  I got most of it from the GitHub docs [1], as
well as looking at other projects.  I'll probably hold off on landing
this until a couple more things land, then land this and tag v0.1.0.

[1] https://docs.github.com/en/communities

Issue: https://github.com/Khan/genqlient/issues/35


benjaminjkraft:
See also: https://github.com/Khan/khan.github.io/pull/17

dangoor:
Nice! Yeah, there are a lot of little bits and pieces, but it does make for a cleaner launch. Thanks!

jvoll (approved):
I've never open-sourced anything before but, this seems good to me!

README.md
```diff
@@ -1,5 +1,10 @@
 <img width="100%" alt="generated graphql client ⇒ genqlient" src="docs/images/genqlient.svg">
 
+[![Go Reference](https://pkg.go.dev/badge/github.com/Khan/genqlient.svg)](https://pkg.go.dev/github.com/Khan/genqlient)
+[![Test Status](https://github.com/Khan/genqlient/actions/workflows/go/badge.svg)](https://github.com/Khan/genqlient/actions)
+[![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)](docs/CODE_OF_CONDUCT.md)
+
```
StevenACoffman, line 6:
```suggestion
[![GoReportcard](https://goreportcard.com/badge/github.com/Khan/genqlient?status.svg)](https://goreportcard.com/report/github.com/Khan/genqlient)

```

StevenACoffman (approved):
One other possible badge. It looks like it's impossible for the test badge to ever change though?

StevenACoffman:
The https://goreportcard.com/report/github.com/Khan/genqlient reports some linting nits which it might be worth addressing to give people more confidence.

benjaminjkraft:
> It looks like it's impossible for the test badge to ever change though?

What do you mean?  I think it would change if, say, someone merged a PR overriding the failing checks.  Or am I missing something that would cause it to stay green even in that case?  (Of course we don't intend to do that, and honestly I think all the badges* are kinda unnecessary, but I can see how it provides some value for people to know that we are actually careful about the things we should be careful about.)

*other than the pkg.go.dev one, which is a useful icon for a necessary link

> The https://goreportcard.com/report/github.com/Khan/genqlient reports some linting nits which it might be worth addressing to give people more confidence.

I fixed the typo (#100).  The rest of the suggestions appear to be:
- unnecessary suggestions from a [deprecated](https://github.com/golang/go/issues/38968) tool, mostly about missing documentation in an internal (and test-only) package
- incorrect suggestions supposedly from `ineffassign`, but they must be from a [broken version](https://github.com/gojp/goreportcard/issues/340) because we run that linter in `golangci-lint`
That doesn't make me think there is significant value to genqlient in suggestions Go Report Card.  I can add the badge, though, since I guess an "A" is a "good" grade even if I fully intend to ignore their suggestions going forward.

StevenACoffman:
Ah great. Thanks!

Issue #85, opened 2021-09-09
Option to use a struct for interface types


benjaminjkraft:
If you request a field of interface type, but your selection has only interface fields (i.e. no fragments), we could generate a simple struct type, instead of the usual interface-and-implementations.  We don't do it by default, because then if you later add fragments you have to change all your code, but it shouldn't be too hard to add an option.

PR #97, opened 2021-09-15
Add a new option to treat an interface like an object

benjaminjkraft:
## Summary:
The basic idea here is if you only request interface fields (no
fragments) you may not care about the concrete type, and so we could
just generate a struct as if it were an object.  I don't think it's a
good idea to do that by default, because then if you later add a
fragment all your code totally changes, but it's quite reasonable as an
option!

Most of the code involved is just wiring and validation; the
core implementation is literally just: treat it like an object.

Issue: https://github.com/Khan/genqlient/issues/85

## Test plan:
make check


docs/FAQ.md
```diff
@@ -183,6 +183,28 @@ if novel, ok := resp.Favorite.(*GetBooksFavoriteNovel); ok {
...
+
+genqlient will generate just:
+
+```go
+type GetBooksFavoriteBook struct {
+  Title string
+}
+```
+
+Keep in mind that if you later want to add fragments to your selection, the types will change entirely.
```
csilvers, line 205: Hmm, if you say `struct: true` and specify fragments, how does it work?  Wouldn't it give an error?

csilvers (commented):
This is straightforward enough I feel comfortable reviewing it -- it looks good to me!

StevenACoffman (approved):
Looks great to me!

docs/FAQ.md
benjaminjkraft: Yeah, I just mean it'll error so you'll have to change it.  Will reword.

PR #98, opened 2021-09-15
Lint against fmt.Println except in main.go

benjaminjkraft:
## Summary:
Usually this is an artifact of debugging, not intentional!  And it's
come up a few times, luckily all noticed by reviewers.

## Test plan:
make check


dnerdy (approved):
🎉

PR #99, opened 2021-09-16
Autofix lint when we can

benjaminjkraft:
## Summary:
This is mostly useful for gofumpt, although a few other linters may
support it.  Anyway, we may as well.

## Test plan:
make lint


dnerdy (approved):

StevenACoffman (approved):
Once golangci-lint supports applying the newer "suggested fixes" it will get a lot more useful.

Issue #38, opened 2021-04-22
Allow mapping directly to a type but not using its standard JSON serialization


benjaminjkraft:
You might want to map your GraphQL `DateTime` to `time.Time`, but you might want to only include microseconds (because Python doesn't like them).  You can do that by defining your own type, but then you have to cast at every call, which is the kind of boilerplate genqlient is designed to reduce!  We could instead let you just define marshal/unmarshal functions, like [gqlgen does](https://gqlgen.com/reference/scalars/#custom-scalars-with-third-party-types).

PR #101, opened 2021-09-16
Refactor import-loading to simplify the type-generation code

benjaminjkraft:
## Summary:
genqlient has some code (`imports.go`) dedicated to tracking which
imports we need and avoiding conflicts, as well as converting a
(restricted) Go expression like `map[string]github.com/me/mypkg.MyType`
to an import (`github.com/me/mypkg`) and a type-reference
(`map[string]mypkg.MyType`) to be used in the context of that import,
and at least making some attempt to track conflicts.  (Right now the
conflict-avoidance is not very smart, and not very well tested, but it
comes up rarely anyway.)  Sadly, that code was a bit cumbersome to use,
because you had to first register the imports (typically from
`convert.go`), then use them (often from the template).

In this commit I refactor the order we write things in order to allow a
significant simplification of how we import; in particular we no longer
have to guess in advance what imports which template will need; it can
just do `{{ref <expr>}}` as before, and it just works.  To do this, I:
- changed the importer to have only one API, which adds the import if
  needed, and returns the reference either way
- added a check that we don't add imports after they're written
- reorganized the toplevel templates a bit to make sure that check never
  fires; we now generate all the types and operations, then write the
  imports and glue it all together
This removes a bunch of silly code, and should simplify the process of
adding custom (un)marshalers (#38).

While I was at it, I put the documentation of what expressions we
support in a more visible place, and added a type-assertion that your
custom context type implements context.Context (if applicable).

## Test plan:
make check


StevenACoffman (approved):
Oh, thanks! The `$` was kind of bugging me too.

generate/generate.go
```diff
@@ -362,24 +353,55 @@ func Generate(config *Config) (map[string][]byte, error) {
...
+	// buffer, since they affect what imports we'll put in the header.
+	var bodyBuf bytes.Buffer
+	err = g.WriteTypes(&bodyBuf)
 	if err != nil {
 		return nil, err
 	}
-	for _, op := range document.Operations {
-		if err = g.addOperation(op); err != nil {
+	for _, operation := range g.Operations {
+		err = g.execute("operation.go.tmpl", &bodyBuf, operation)
```
dnerdy, line 375: I know "execute" is based on text/template terminology, but that isn't super clear in the context of the generator, IMO. Perhaps this function could be named `renderTemplate` or something similar.

generate/imports.go
```diff
@@ -80,22 +66,20 @@ func (g *generator) getRef(fullyQualifiedName string, addImport bool) (qualified
...
-		var ok bool
-		alias, ok = g.imports[pkgPath]
-		if !ok {
-			// This is an internal error, not a user error.
-			return "", errorf(nil, `no alias defined for package "%v"`, pkgPath)
+	alias, ok := g.imports[pkgPath]
+	if !ok {
+		if g.importsLocked {
+			return "", errorf(nil,
+				`genqlient internal error: imports locked but no alias defined for package "%v"`, pkgPath)
```
dnerdy, line 73: perhaps: "imports locked but package "%v" has not been imported"?

dnerdy (approved):
I like it! The buffering is much easier to reason about compared to pre-registering references.

Issue #43, opened 2021-05-03
omitempty generates broken code for structs


benjaminjkraft:
You get
```go
	var zero_myVar []string
	if myVar != zero_myVar {
		variables["myVar"] = myVar
	}
```
which is semantically correct but doesn't compile, because arrays aren't comparable.  We should instead compare to `nil` (or check `len`, depending what we want to do with `[]string{}`) for slices and maps.

As a workaround, since `[]string(nil)` encodes to `null`, there's usually no need for `omitempty` on such fields, unless the backend actually makes a distinction between the argument being null and unset.

benjaminjkraft:
This also affects `struct { F []string }`, which will be a little harder to fix.

benjaminjkraft:
This is also incorrect if you use `bindings` to map some type to a slice.

PR #103, opened 2021-09-17
Refactor argument-handling to use a struct

benjaminjkraft:
## Summary:
In this commit I refactor the argument-generation logic to move most of
the code out of the template and into the type-generator.  This logic
predates #51, and I didn't think to update it there, but I think it
benefits from similar treatment, for similar reasons.

Specifically, the main change is to treat variables as another struct
type we can generate, rather than handling them inline as a
`map[string]interface{}`.  Users still pass them the same way, but
instead of putting them into a `map[string]interface{}` and JSONifying
that, we generate a struct and put them there.

This turns out to simplify things quite a lot, because we already have a
lot of code to generate types.  Notably, the omitempty code goes from a
dozen lines to basically two, and fixes a bug (#43) in the process,
because now that we have a struct, `json.Marshal` will do our work for
us! (And, once we have syntax for it (#14), we'll be able to handle
field-level omitempty basically for free.)  More importantly, it will
simplify custom marshalers (#38, forthcoming) significantly, since we do
all that logic at the containing-struct level, but will need to apply it
to arguments.

It does require two breaking changes:

1. For folks implementing the `graphql.Client` API (rather than just
   calling `NewClient`): we now pass them variables as an `interface{}`
   rather than a `map[string]interface{}`.  For most callers, including
   Khan/webapp, this is basically a one-line change to the signature of
   their `MakeRequest`, and it should be a lot more future-proof.
2. genqlient's handling of the `omitempty` option has changed to match
   that of `encoding/json`, in particular it now never considers structs
   "empty".  The difference was never intentional (I just didn't realize
   that behavior of `encoding/json`); arguably our behavior was more
   useful but I think that's outweighed by the value of consistency with
   `encoding/json` as well as the simpler and more correct
   implementation (fixing #43 is actually quite nontrivial otherwise).
   Once we have custom unmarshaler support (#38), users will be able to
   map a zero value to JSON null if they wish, which is mostly if not
   entirely equivalent for GraphQL's purposes.

Issue: https://github.com/Khan/genqlient/issues/38
Issue: https://github.com/Khan/genqlient/issues/43

## Test plan:
make check

### DIFF
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -180,6 +180,7 @@ func (g *generator) convertArguments(
 		GoName:    name,
 		Fields:    fields,
 		Selection: nil,
+		IsInput:   true,
 		descriptionInfo: descriptionInfo{
 			CommentOverride: fmt.Sprintf("%s is used internally by genqlient", name),
 			// fake name, used by addType
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -7,7 +7,7 @@ package generate
 // into code, in types.go.
 //
 // The entrypoints are convertOperation, which builds the response-type for a
-// query, and convertInputType, which builds the argument-types.
+// query, and convertArguments, which builds the argument-types.
 
 import (
 	"fmt"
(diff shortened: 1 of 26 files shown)


benjaminjkraft:
Hmm, I realized this actually makes some more breaking changes, because my implementation of `omitempty` had some inconsistencies with JSON around structs.  Let me think about this some more.

StevenACoffman (approved):
So that was my concern... that it is "breaking"-ish, which I guess if fine before our blog post and initial release version, but I'd like to do that sooner rather than later, and be more sure we don't need to again if we do this.

benjaminjkraft:
Yeah, I think if we had more users I'd be more inclined to live with the existing behavior, but now's a good time to fix the inconsistency before anyone cares.  (FWIW, this appears to be something that `yaml` changed from v1 to v2!)  In practice it's not hard to search for potential problems (just find struct-typed fields with `# genqlient(omitempty: true)`), and webapp has none.

generate/convert.go
```diff
@@ -139,15 +139,64 @@ var builtinTypes = map[string]string{
...
+	for i, arg := range operation.VariableDefinitions {
+		_, directive, err := g.parsePrecedingComment(arg, arg.Position)
+		if err != nil {
+			return nil, err
+		}
+		options := queryOptions.merge(directive)
+
+		goName := upperFirst(arg.Variable)
+		// note prefix is ignored here (see generator.typeName), as is
+		// selectionSet (for input types we use the whole thing).
```
dnerdy, line 165: I agree with the comment in the code below that it's a bit confusing to pass around arguments that aren't used. It may be helpful to explicitly say that selection sets don't apply to input objects, and input types may only be scalars, enums or (recursively) other input objects. A mention of ast.InputObject in convertDefinition might also be helpful.

Also, I think generator.typeName is a dangling reference? (Here and below in convertDefinition.)

dnerdy (approved):
Nice. I don't feel like we need to worry about breaking anyone at this point, plus the change is well documented.

PR #104, opened 2021-09-17
Add support for binding with a custom marshal/unmarshal function

benjaminjkraft:
## Summary:
This is useful if you want to bind to a type you don't control (or use
for other things) but need different serialization than its default.
This is a feature gqlgen has and we've found it very useful.  For
example, in webapp we want to bind `DateTime` to `time.Time`, but its
default serialization is not compatible with Python, so currently we
have to bind to a wrapper type and cast all over the place, which is
exactly the sort of boilerplate genqlient is supposed to avoid.

For unmarshaling, the implementation basically just follows the existing
support for abstract types; instead of calling our own generated
helper, we now call your specified function.  This required some
refactoring to abstract the handling of custom unmarshalers generally
from abstract types specifically, and to wire in not only the
unmarshaler-name but also the `generator` (in order to compute the right
import alias).

For marshaling, I had to implement all that stuff over again; it's
mostly parallel to unmarshaling (and I made a few minor changes to
unmarshaling to make the two more parallel).  Luckily, after #103 I at
least only had to do it once, rather than implementing the same
functionality for arguments and for input-type fields.  It was still
quite a bit of code; I didn't try to be quite as completionist about the
tests as with unmarshal but still had to add a few.

Issue: https://github.com/Khan/genqlient/issues/38

## Test plan:
make check


StevenACoffman (approved):
Yay! Great!

graphql/util.go
```diff
@@ -10,10 +10,18 @@ package graphql
 //	type T struct { E; NoUnmarshalJSON }
 // where E has an UnmarshalJSON method, T will not inherit it, per the Go
 // selector rules: https://golang.org/ref/spec#Selectors.
+//
+// It also works for marshaling, in the same way (despite the name).
```
dnerdy, line 14: What do you think about naming this NoMarshalJSON? It seems a little more expected that a type with this name would also handle unmarshaling vs. the other way around. (Another option is to have a completely separate type so marshalnig and unmarshaling can be handled separately.)

dnerdy (approved):
Looks good to me! This definitely adds a bit of additional complexity, but it also seems like it will be really useful.

graphql/util.go
benjaminjkraft: I'm not super excited to rename it, but maybe I'll just add a separate NoMarshalJSON; we only ever need one.

PR #105, opened 2021-09-17
Fix broken behavior for several options on the same node

benjaminjkraft:
## Summary:
Previously, we actually allowed you to put several genqlient directives
on the same node, but the semantics were undocumented (and somewhat
confusing, when it comes to `typename`).  In order to support directives
on input options, we're actually going to be encouraging this usage (see
notes in #14), so it's time to fix it.  To avoid confusion, I just had
conflicting directives be an error, rather than defining which one
"wins".  The same applies to specifying the same option several
times in one directive.

I also fixed two small bugs:
- `typename` on an operation would incorrectly cascade down to
  all input types in a query (causing conflicts).
- directive parse errors had useless positions, now they're correct

## Test plan:
make check


StevenACoffman (approved):
Oh, thanks!

docs/genqlient_directive.graphql
```diff
@@ -21,15 +21,25 @@
...
 #  }
 # the directive "a" is ignored, "b" and "c" apply to all relevant nodes in the
-# query, "d" applies to arg2 and arg3, and "e" applies to field1 and field2.
+# query, "d" applies to arg2 and arg3, "e" applies to field1 and field2, and
+# "f" applies to field3.
+#
+# Except as noted below, directives on nodes take precedence over ones on the
+# entire query (so "d", "e", and "f" take precedence over "b" and "c"), and
+# multiple directives on the same node ("b" and "c") must not conflict.  Note
+# that directives on nodes do apply to their "children", so "d" does not apply
```
dnerdy, line 41: I think this means to say: Note that directives on nodes do NOT apply to their "children".

dnerdy (approved):

Issue #109, opened 2021-09-22
Type-name collapsing not working in some cases


benjaminjkraft:
When pulling `v0.1.0` into webapp we got some changes to type names which I think must have been inadvertently introduced by #71.  My guess is something about where we do the uppercasing is confusing the suffix-collapsing logic.  The details are in https://phabricator.khanacademy.org/D74434.

PR #110, opened 2021-09-22
Fix a bug in type-name collapsing introduced by #71

benjaminjkraft:
## Summary:
We typically name our types `OperationFieldTypeFieldType`, but if a
type's name matches the preceding field-name, we omit the type-name.
In #71 I changed the behavior such that we no longer do that in the case
where the type's name matches some suffix of the name-so-far that's
longer than just the leaf field-name.

This was semi-intentional; I assumed it didn't matter and would be more
predictable this way.  But it turns out that was a feature, both in the
sense that almost any change to the type-name-generator is breaking, and
in the sense that it made the names uglier.  Plus, now that we have
better conflict-detection (#94), the possibility that some tricksy
type-names could cause problems is no longer as much of an issue, so we
can be a little less careful here.  (Although I think this is no less
safe than before; the field-names are the important part.)  So in this
commit I revert the change.

Specifically, this comes up a lot at Khan where we do
```
mutation ForcePhantom {
    forcePhantom {     # type: ForcePhantom
        error { ... }  # type: ForcePhantomError
    }
}
```
Before #71, and again after this change, we'll generate
`ForcePhantomForcePhantomError` for `error`; before we'd generate
`ForcePhantomForcePhantomErrorForcePhantomError`.

Issue: https://github.com/Khan/genqlient/issues/109

## Test plan:
make tesc

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -26,6 +26,8 @@ When releasing a new version:
 
 ### Bug fixes:
 
+- Generated type-names now abbreviate across multiple components; for example if the path to a type is `(MyOperation, Outer, Outer, Inner, OuterInner)`, it will again be called `MyOperationOuterInner`.  (This regressed in a pre-v0.1.0 refactor.) (#109)
+
 ## v0.1.0
 
 First open-sourced version.
--- a/generate/names.go
+++ b/generate/names.go
@@ -143,9 +143,9 @@ func typeNameParts(prefix *prefixList, typeName string) *prefixList {
 	// If the prefix has just one part, that's the operation-name.  There's no
 	// need to add "Query" or "Mutation".  (Zero should never happen.)
 	if prefix == nil || prefix.tail == nil ||
-		// If the last prefix field ends with this type's name, omit the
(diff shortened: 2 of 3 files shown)


csilvers (approved):
OK!  I'm not the best person to review, but it's a small enough change, and I think the review is in the semantics really, which seem fine to me.

csilvers:
This is merged into `main` but I'm not seeing any differences. I ran
```
   go get github.com/Khan/genqlient@main
```
which updated it to `v0.1.1-0.20210917010932-f72933fa0e0b` which seems plausible, but when I re-run `make genqlient` I don't see any type-names change.

csilvers:
(Ah, nope, f729 is not the latest commit.  I wonder what I have to do to get `go get` to get the latest.)

csilvers:
Ah, I got it by doing `go get github.com/Khan/genqlient@ab...` directly.

benjaminjkraft:
Possibly just a caching issue?  Specifying the commit should work anyway.

dnerdy:
The change looks good to me.

Issue #30, opened 2021-04-22
Collapse trivial structs


benjaminjkraft:
If you do `mutation { MyMutation { error { code } } }`, or something, we should be able to just return `(code string, err error)`, at least if you ask us too.  This would be a nice convenience -- the helper's callers no longer have to know as much about the query structure -- but the rules and logic to do it are surprisingly subtle.

benjaminjkraft:
This would be especially useful for named fragments, where you might do
```graphql
f { ...MyFragment }
```
and it would be great if we could just generate a field `F MyFragment` instead of (effectively) `F struct { MyFragment }`.

benjaminjkraft:
This has come up quite a lot with fragments.  I'm going to try to work on it if I get time.

benjaminjkraft:
#121 implements this, but only for fragment spreads, as they've proven to be both the more useful case and by far the easier one to implement.  See the PR for details.

Issue #108, opened 2021-09-22
Support for tests/mocking in genqlient


csilvers:
My thought here is that for every function `Myquery` that genqlient produces code for, it also produces a function `MockMyquery`, that can be used in tests.  But we do away with strong typing altogether!, and use the js.Obj system from Khan Academy.  The justification for this is that, unlike for prod code, it's ok to do type-checking at runtime for tests.  And specifying js.Obj{} objects is just much nicer to write than all the super-long type-names that would be provided if we specified Go types.

Specifically, my thought is the API might be:
    MockMyquery(ctx context.Context, testClient graphql.Client, varsToMatch js.Obj, output js.Obj)
and this would add a mux to testClient so whenever you queried that client on Myquery with args that match `varsToMatch`, the client would return `output`.  (Alternately, we could have you pass in a mux yourself, and you could attach it to your test-client however you want; that matches the httplib model better and may be easier or harder to use, I'm not sure.)

The logic of MockMyquery() would include code to validate the schema: to make sure that varsToMatch includes all required vars and they can all be coerced to the proper type, and likewise that the output matches the query's schema.

varsToMatch could include all the js matchers that we have at Khan, such as `js.UnorderedArray{}`.  It could also include a new `js.Any` to match anything.  (Because leaving a required var out of varsToMatch would be an error, rather than the existing Khan Academy logic where it's an implicit js.Any.)  You could just pass in `js.Any` directly for `varsToMatch` to mean "return the given output for all Myquery queries."

I am thinking the codegen would also just create the test helper files to a package generated/genqlient/js/.  So clients could import that and use js.Obj/etc like they do at Khan Academy already.

[genqlient.tar.gz](https://github.com/Khan/genqlient/files/7212703/genqlient.tar.gz) holds the relevant files we use at Khan Academy:
* The files that would go in the genqlient/js/ dir (in the js/ dir in the tarball)
* mux.go (in case it's helpful)
* query_validation.go, response_validation.go (how we validate against the schema)
* testing.go (to create a mock client)

benjaminjkraft:
Thanks for the thoughts!  I'm gonna put some time on the calendar with you and @dnerdy to think through some of this synchronously.  In the meantime, some less-organized thoughts.

[...]

PR #111, opened 2021-09-23
Include __typename on interface-typed fragments

benjaminjkraft:
## Summary:
We ask for `__typename` on interface-typed fields so that we know which
type to use when unmarshaling.  We never strictly need this for
fragments, because the context into which they're spread must either
have `__typename` or must be of concrete type.  But a few cases have
come up where it would be somewhat convenient: first, if you're using
genqlient's types for mocking (not really supported right now, but
possible); and second, for consistency if we add an option to "flatten"
fragments (#30), which would effectively drop the `__typename` field
from the structs otherwise.  Neither is a deeply compelling reason, I
feel, but it's not too hard to do so we may as well.

## Test plan:
make check


StevenACoffman (approved):

dnerdy:
I'm not entirely opposed to this change, but it does seem unnecessary. The mock response validation issue was an issue with the way the validator was handling abstract spreads in a concrete context. That's been fixed. That said, if this has utility for mocking in other ways, perhaps we do want it.

benjaminjkraft:
Yeah, there are two ways this is useful:
1. It lets you be a little less careful if you're mixing `js.Obj` and genqlient types in mocks and don't want to match the full genqlient structure exactly.  In practice that's used as a workaround for the fact that marshaling genqlient (output) types doesn't work right with embeds.
2. Once we add an option to "flatten" selections that are just a single fragment-spread (#30), there will be no typename field in the final response struct.  This is fine for genqlient -- it only needs the field in GraphQL and including it in structs was sort of accidental.  But it has proven useful for mocking, if your mock needs to set `__typename`; and it could be helpful for humans to be able to call the method too (although that's hypothetical).
We can talk more this afternoon at our meeting about #108.

benjaminjkraft:
This is on hold for now to see if we need it (see comments on #108).

benjaminjkraft:
This is now replaced by #120, which writes `__typename` when marshaling without any need for the user to fill in a field.

PR #112, opened 2021-09-24
[🔥AUDIT🔥] Fix tests after merge

benjaminjkraft:
🖍 _This is an audit!_ 🖍

## Summary:
This is what I get for not waiting for them to pass at each step of
landing a stack!

## Test plan:
make check

PR #114, opened 2021-09-25
[🔥AUDIT🔥] Remove security issue template

benjaminjkraft:
🖍 _This is an audit!_ 🖍

## Summary:
I figured we should have this to say "STOP! Use HackerOne instead".
But GitHub anticipated that and already links to the security policy
(`docs/SECURITY.md`) as an option when [creating an issue][1], so
it's just duplicative.  In this commit I remove it.

[1]: https://github.com/Khan/genqlient/issues/new/choose

## Test plan:
above link should show just one option for security


benjaminjkraft:
Note checks failed on this commit but are fixed by #112 which I just forgot to merge in this branch (but is in `main`).

Issue #47, opened 2021-05-20
Use a codegen-based JSON library


benjaminjkraft:
Any code that's using genqlient should be a great match for one of the libraries that does JSON via codegen rather than reflection, which are theoretically way better for perf.  @StevenACoffman pointed me to [easyjson](https://github.com/mailru/easyjson), [ffjson](https://github.com/pquerna/ffjson), [jsonparser](https://github.com/buger/jsonparser), and [gojay](https://github.com/francoispqt/gojay), but I haven't looked at any of them.

We do need to make sure this won't cause trouble with custom scalars: it might be a little surprising if you have to rerun genqlient because you changed a json tag in a custom scalar, although it's not a huge problem in practice I expect (especially since probably most custom scalars will have `UnmarshalJSON` methods).  In any case we could also put this behind a flag.

benjaminjkraft:
As mentioned in #120, another option is to roll our own, making use of what we know about our types.  I suspect it's not worth it yet, but if our (un)marshalers start having to get even more complicated, it may be, since we end up doing a lot of work to get `encoding/json` to do what we want.

csilvers:
go-codec also supports codegen for json (and other encoding formats as well) I believe, via `codecgen`.

StevenACoffman:
[ugorji/go](https://github.com/ugorji/go) is worth adding to the list above. It is [described here](http://ugorji.net/blog/go-codecgen).
Supported Serialization formats are:

  - msgpack: https://github.com/msgpack/msgpack
  - binc:    http://github.com/ugorji/binc
  - cbor:    http://cbor.io http://tools.ietf.org/html/rfc7049
  - json:    http://json.org http://tools.ietf.org/html/rfc7159
  
I hadn't mentioned it before, because I'm not sure that anything but json is relevant to us, and so it's possible part of [ugorji/go](https://github.com/ugorji/go) popularity is only due to non-relevant formats. However, if it's well maintained, and has the right features, it's worth considering.
 
| library |  order by page rank | number of github stars | 
|---|---|---|
| [go-codec](https://www.grank.io/pkg/github.com/ugorji/go/codec%20(github.com/ugorji/go).html) | 145 | 1371 |
| [github.com/mailru/easyjson ](https://www.grank.io/pkg/github.com/mailru/easyjson.html) | 185 | 3331 |
| [github.com/buger/jsonparser](https://www.grank.io/pkg/github.com/buger/jsonparser.html) | 338  | 4185 |
[...]

PR #120, opened 2021-09-29
Allow genqlient types to be marshaled safely

benjaminjkraft:
## Summary:
When genqlient generates output types, it generates whatever code is
necessary to unmarshal them.  Conversely, when it generates input types,
it generates whatever code is necessary to marshal.  This is all that's
needed for genqlient itself: it never needs to marshal output types or
unmarshal input types.

But maybe you do!  (For example, to put the responses in a cache, which
is the use case that @csilvers hit at Khan, although there are others
one can imagine.)  While we can't support every serialization format you
might want (at least not without adding plugins or some such), it's not
unreasonable to expect that since genqlient can read JSON, it can write
it too.  Sadly, in the past this was not true for types requiring custom
unmarshaling logic, for several reasons.

In this commit I implement logic to always write both marshalers and
unmarshalers whenever they're needed to be able to correctly round-trip
the types, even though genqlient doesn't do so.  I wasn't starting from
scratch, since of course we already write both marshalers and
unmarshalers in some cases.  But this ended up requiring surprisingly
large changes on the marshaling side, mostly to correctly support
embedding (which we use for named fragments).

Specifically, as the comments in `types.go` discuss, the most difficult
issue is spreads with duplicate fields, which translate to Go embedded
fields which end up hidden from the json-marshaler.  Ultimately, I had
to do things quite differently from unmarshaling, and essentially
flatten the type when we write marshaler.  But in the end it's not so
ugly -- indeed arguably it's cleaner!  Mainly it's just different.

One thing to note is that we do marshal `__typename` based on
what we know about the types; users need not fill it in (and if they
do we'll ignore it).  This seemed to me to be a better UX, and
didn't add much complexity.

In general, I begin to wonder whether using `encoding/json` at all is
really right for genqlient: we're doing a lot of work to appease it,
despite knowing what our types look like.  I think it would still be a
significant increase in lines of code to roll our own, but that code
would perhaps be simpler, and would surely be faster (although if we
just want the speed gains we could use another JSON-generator library,
see also #47).  Anyway, something to think about in the future.

## Test plan:
make tesc

### DIFF
--- a/generate/types.go
+++ b/generate/types.go
@@ -158,6 +158,17 @@ func (field *goStructField) IsEmbedded() bool {
 	return field.GoName == ""
 }
 
+// Selector returns the field's name, which is unqualified type-name if it's
+// embedded.
+func (field *goStructField) Selector() string {
+	if field.GoName != "" {
+		return field.GoName
+	}
+	// TODO(benkraft): This assumes the type is package-local, which is always
+	// true for embedded types for us, but isn't the most robust assumption.
+	return field.GoType.Unwrap().Reference()
+}
+
 // unmarshaler returns:
 // - the name of the function to use to unmarshal this field
 // - true if this is a fully-qualified name (false if it is a package-local
@@ -176,14 +187,6 @@ func (field *goStructField) unmarshaler() (qualifiedName string, needsImport boo
 	return "encoding/json.Unmarshal", true, field.IsEmbedded()
 }
 
-// NeedsUnmarshaler returns true if this field needs special handling when
-// unmarshaling, e.g. if it's of interface type, embedded, or has a
-// user-specified custom unmarshaler.
-func (field *goStructField) NeedsUnmarshaler() bool {
-	_, _, ok := field.unmarshaler()
-	return ok
-}
-
 // Unmarshaler returns the Go name of the function to use to unmarshal this
 // field (which may be "json.Unmarshal" if there's not a special one).
 func (field *goStructField) Unmarshaler(g *generator) (string, error) {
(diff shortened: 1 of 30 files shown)


generate/types.go
```diff
@@ -198,34 +201,149 @@ func (field *goStructField) Unmarshaler(g *generator) (string, error) {
...
+// According to Go's embedding rules, QT has no field Id: since QT.A.Id and
+// QT.B.Id are at equal depth, neither wins and gets promoted.  (Go's JSON
+// library uses similar logic to decide which field to write to JSON, except
+// with the additional rule that a field with a JSON tag wins over a field
+// without; in our case both have such a field.)
+//
+// Those rules don't work for us.  When unmarshaling, we want to fill in all
+// the potentially-matching fields (QT.A.Id and QT.B.Id in this case), and when
+// marshaling, we want to always marshal exactly one potentially-conflicting
+// field; we're happy to use the Go visibility rules when they apply. such field, choosing an
```
csilvers, line 280: I think the text got a bit garbled here.

internal/testutil/types.go
```diff
@@ -31,6 +31,9 @@ func GetClientFromMyContext(ctx MyContext) (graphql.Client, error)     { return
 const dateFormat = "2006-01-02"
 
 func MarshalDate(t *time.Time) ([]byte, error) {
+	if t == nil || t.IsZero() {
+		return []byte("null"), nil
+	}
```
csilvers, line 36: Why did you add this back in?

csilvers:
I'm probably not the right person to review this, but it all seems plausible to me!  It is indeed tricky stuff, but it's commented really well.

I do feel like the code would be easier to understand in some ways, actually, if you codegen-ed the json marshal/unmarshal code; you're kinda doing that already for the hard cases.  But I agree it would make the size of these generated files explode -- even for the common (?) use-cases where people don't have these fancy embeds and interfaces -- which is a consideration in itself.

StevenACoffman (approved):
This took me a while to read through, and I think it's great, and we should merge it. I did wonder if we could offload some of your heavy lifting to a library... but other than possibly the one I slacked about, I didn't casually find one that was an obvious candidate.

internal/testutil/types.go
benjaminjkraft: The `t == nil` should be unnecessary, I just put it there for completeness.  But `t.IsZero()` we do want, to avoid marshaling a `time.Time` field to `"0001-01-01"`, since genqlient matches `encoding/json` in counting zero structs as nonempty for purposes of `omitempty`.

benjaminjkraft:
> I do feel like the code would be easier to understand in some ways, actually, if you codegen-ed the json marshal/unmarshal code; you're kinda doing that already for the hard cases.

Yeah, that's my theory.  I think there's a lot of stuff to write to handle things like `omitempty` as well as the basics of encoding strings, bools, numbers, etc., all of which we're avoiding now, so I think even just for the genqlient code it would be a big increase in the quantity of code.  But I think it would all be much simpler.  I may prototype it if I'm bored sometime, and see how it looks.

> But I agree it would make the size of these generated files explode -- even for the common (?) use-cases where people don't have these fancy embeds and interfaces -- which is a consideration in itself.

Yeah, I don't know how much of an issue that will be for folks.  It's probably a nonzero perf/binary-size impact, too, in that I don't think the Go compiler/linker can be very smart about pruning methods since you might (and indeed `encoding/json` does) call them via reflection.  This PR already adds quite a bit; I think if it becomes an issue for anyone we could add an option to go back to the old behavior of generating marshalers only for input types and unmarshalers only for output types.  I think generating both is the better "batteries-included" default; people who care a lot about binary size should expect to need to tune some things.  But with generating the json-encoding code, we wouldn't want to have to support both ways.

> I did wonder if we could offload some of your heavy lifting to a library... but other than possibly the one I slacked about, I didn't casually find one that was an obvious candidate.

Yeah, thanks for raising that option.  I was thinking about how I feel I am learning more and more places where the stdlib API leaves a lot to be desired, but I didn't think to actually look for what others have done.  Sadly, as I wrote on #120, as far as I can tell none of the ones we've found will help a lot with the marshaling side, where embedding is the real challenge, rather than interfaces and custom marshalers.  (I mean, both require nontrivial code, but the latter is just copied from unmarshaling, so if you went to the pains to wrap your head around what we do there, it's not so bad.)  My feeling is that on the whole it doesn't make sense (at least not just for the code-complexity gains; if we used go-codec the perf benefits might be worth it).  But I think in any case since the biggest benefits are on the unmarshaling side, I'd want to do it as a separate PR.

dnerdy (approved):
The two-pass marshaling process is an interesting solution! A custom marshaler does seem like it could simplify the code, and that would be a good approach to explore.

PR #124, opened 2021-10-01
Add a mechanism to specify options on input-type fields

### LINKED ISSUE #123: Validate against conflicting options on the same type
Consider the following query:
```graphql
query Q(
  # @genqlient(omitempty: true)
  input1: MyInput,
  input2: MyInput,
) { ... }
```

In this case, we generate a type `MyInput` which we use for both `input1` and `input2`.  But there's actually a conflict here: `omitempty` is set to true on `input1` and false (by default) on `input2`.  The same problem is even more likely to happen with different queries in a package, and also applies to options set via the forthcoming input-field options setting (see #14).

Right now we don't check for that; we just use whichever one we see first.  Instead, similar to how we validate that if you use `typename` on several fields, they all have the same selections, we should validate that in any case where we use the same type-name in two places, they have the same options.  So in this case, we would error, and advise the user to either set `omitempty: true` on `input2`, or set a different typename for either `input1` or `input2`.

In principle we could do this by extending the existing naming-collision validation (`selectionsMatch`).  But doing that in the obvious way will require some annoying plumbing as well as parsing directives twice (once in type-generation then again in validation).  The best thing to do is probably to parse all the directives in a preprocessing step upfront rather than as we generate types, and somehow attach them to nodes (perhaps just via a global map of node to genqlient directive).  Then we can just read from that map in both type-generation and validation.  (It would probably simplify the conversion code, too, since we could skip plumbing the options everywhere inline.)

Note that this will be a breaking change, but it will be going from silent confusing behavior to an explicit error, so I think that's totally fine and worthwhile.

benjaminjkraft:
## Summary:
This has been a bit of a thorn since we started using genqlient in
production: just as you might want to specify, say, `omitempty` on an
argument, you might equally want to specify it on an input-type field.
But there's no obvious syntax to do that, because the input-type field
does not appear in the query (only the schema) so there's nowhere to put
the `# @genqlient` directive.

This commit, at last, fixes that problem, via a new option, `for`, which
you use in an option applied to the entire operation (or fragment), and
says, "actually, apply this directive to the given field, not the entire
operation".  (It's mainly useful for input types, but I allowed it for
output types too; I could imagine it being convenient if you want to say
you always use a certain type or type-name for a certain field.)  It
works basically like you expect: the inline options take precedence over
`for` take precedence over query-global options.

The implementation was fairly straightforward once I did a little
refactoring, mostly in the directive-parsing and directive-merging
(which are now combined, since merging is now a bit more complicated).
With that in place, and extended to support `for`, we need only add the
same wiring to input-fields that we have for other places you can put
directives.  I did not attempt to solve the issue I've now documented
as #123, wherein conflicting options can lead to confusing behavior;
the new `for` is a new and perhaps more attractive avenue to cause it
but the issue remains the same and requires nontrivial refactoring
(described in the issue) to solve.  (The breakage isn't horrible for the
most part; the option will just apply, or not apply, where you don't
expect it to.)

But while applying that logic, I noticed a problem, which is that we
were inconsistently cascading operation-level options down to
input-object fields.  (I think this came out of the fact that initially
I thought to cascade them, then realized that this could cause problems
like #123 and intended to walk them back, but then accidentally only
"fixed" it for `omitempty`.  I guess until this change, operation-level
options were rare enough, and input-field options messy enough, that no
one noticed.)  So in this commit I bring things back into consistency,
by saying that they do cascade: with at least a sketch of a path forward
to solve #123 via better validation, I think that's by far the clearest
behavior.

Issue: https://github.com/Khan/genqlient/issues/14

## Test plan:
make check


generate/testdata/queries/Pointers.graphql
```diff
@@ -1,11 +1,14 @@
 # @genqlient(pointer: true)
+# @genqlient(for: "UserQueryInput.id", pointer: false)
+# @genqlient(for: "User.id", pointer: false)
 query PointersQuery(
```
csilvers, line 4: I think it would be good to put those example in the docs. All your examples have just one directive and I think it's not necessarily obvious how they combine.   This examples makes it very clear.

generate/genqlient_directive.go
```diff
@@ -77,7 +118,37 @@ func (dir *genqlientDirective) add(graphQLDirective *ast.Directive, pos *ast.Pos
 		return errorf(pos, "the only valid comment-directive is @genqlient, got %v", graphQLDirective.Name)
 	}
 
+	// First, see if this directive has a "for" option;
+	// if it does, the rest of our work will operate on the
+	// appropriate place in FieldDirectives.
 	var err error
+	forField := ""
+	for _, arg := range graphQLDirective.Arguments {
+		if arg.Name == "for" {
```
csilvers, line 127: Do you want to check if `for` is specified multiple times, and error out if so?

generate/genqlient_directive.go
```diff
@@ -77,7 +118,37 @@ func (dir *genqlientDirective) add(graphQLDirective *ast.Directive, pos *ast.Pos
...
+		forParts := strings.Split(forField, ".")
+		if len(forParts) != 2 {
+			return errorf(pos, `for must be of the form "MyType.myField"`)
+		}
+		typeName, fieldName := forParts[0], forParts[1]
+
+		fieldDir := newGenqlientDirective(pos)
+		if dir.FieldDirectives[typeName] == nil {
+			dir.FieldDirectives[typeName] = make(map[string]*genqlientDirective)
+		}
```
csilvers, line 144: Can this happen?  I guess it doesn't hurt to be defensive, but if all genqlientDirectives are made via newGenqlientDirective then this `if` will never be true.

generate/genqlient_directive.go
```diff
@@ -244,68 +361,116 @@ func validateFlattenOption(
...
+// Note this has slightly different semantics than .add(), see inline for
+// details.
+//
+// parent is as described in parsePrecedingComment.
+func (dir *genqlientDirective) mergeOperationDirective(
+	node interface{},
+	parent *ast.Definition,
+	operationDirective *genqlientDirective,
+) {
+	var forField *genqlientDirective
```
csilvers, line 403: This is the `for` field for some parent element whose `for` points to us, is that right?

generate/genqlient_directive.go
```diff
@@ -244,68 +361,116 @@ func validateFlattenOption(
...
 // none), the remaining human-readable comment (or "" if there is none), and an
 // error if the directive is invalid.
+//
+// queryOptions are the options to be applied to this entire query (or
+// fragment); the local options will be merged into those.  It should be nil if
+// we are parsing the directive on the entire query.
+//
+// parent need only be set if node is an input-type field; it should be
+// the type containing this field.  (We can get this from gqlparser in other
+// cases, but not input-type fields.)
```
csilvers, line 440: Do you maybe want to call it parentIfInputField then?

csilvers (approved):
It turns out I need this for one of the mutations I'm converting!  So I'm glad you managed to squeeze it in.  Even though I'm not super-familiar with the genqlient code this seemed self-contained enough that it made sense to me, so I'm happy to approve it!

StevenACoffman (approved):
It turns out I need this for a mutation I'm doing as well, so very timely!

generate/genqlient_directive.go
benjaminjkraft: Yes -- since it's a map of maps we can only initialize the toplevel one upfront.  (It really pains me that Go didn't make the zero value more usable here, but they didn't.)

generate/genqlient_directive.go
benjaminjkraft: Yes, I'll comment better.

generate/genqlient_directive.go
csilvers: Ah, I see.  Yeah, being forced to call `make` to make maps is pretty annoying.  You don't have to do it for slices!, so...

PR #127, opened 2021-10-01
Add the version of genqlient to the generated code

benjaminjkraft:
## Summary:
Craig suggested this; it seems plausibly useful although of course it
will cause some churn in the generated file as well.  And it's easy to
do from Go's automatically added module metadata.  Luckily, something
about how that works means that even when you're on a locally-tagged
commit, in tests the version is `(devel)`, so hopefully this won't cause
any problems for them.

## Test plan:
make check


benjaminjkraft:
@StevenACoffman, I figured I'd put this up as a PR to discuss since it's so easy, but I wanted to check: do you think this is a good idea, a bad idea, or what?  I don't know what's standard here; I'm not attached to this change but if it seems helpful I'm happy to merge it.

benjaminjkraft:
Oy, apparently it doesn't work right in the github action.  I'll have to see if I can find a way to override that to work right.

StevenACoffman:
The test failures is just that the expectations and the results are a little of sync:
```
        // Code generated by github.com/Khan/genqlient version <unknown>, DO NOT EDIT.
```
vs:
```
        // Code generated by github.com/Khan/genqlient version (devel), DO NOT EDIT.
```

StevenACoffman:
It's a fairly common thing. For instance, https://github.com/deepmap/oapi-codegen/ puts:

```
// Package api provides primitives to interact with the openapi HTTP API.
//
// Code generated by github.com/deepmap/oapi-codegen version (devel) DO NOT EDIT.
```

generate/config.go
```diff
@@ -36,6 +37,10 @@ type Config struct {
 	// The directory of the config-file (relative to which all the other paths
 	// are resolved).  Set by ValidateAndFillDefaults.
 	baseDir string
+
+	// The version of genqlient, from runtime.ReadBuildInfo, included in the
+	// generated files.  Set by ValidateAndFillDefaults.
+	Version string
```
StevenACoffman, line 43: You can initialize this to a default value of `(default)` and still it be overridden at build time by the linker flags. 

I often do:
```
// 	Populated during build
var (
	AppName      = "unknown" // What the app is called
	Project      = "unknown" // Which GCP Project e.g. khan-internal-services
	Date         = ""        // Build Date in RFC3339 e.g. $(date -u +"%Y-%m-%dT%H:%M:%SZ")
	GitCommit    = "?"       // Git Commit sha1 of source
	Version      = "v0.0.0"  // go mod version: v0.0.0-20200214070026-92e9ce6ff79f
	HumanVersion = fmt.Sprintf(
		"%s %s %s (%s) on %s",
		AppName,
		Project,
		Version,
		GitCommit,
		Date,
	)
)
```

StevenACoffman (approved):
Seems valuable for troubleshooting, but incomplete. 

What is a "Version"? What is the correct "Version"? This can get tricky.
You can find the most recent tag in this repository using: `git describe --tags $(git rev-list --tags --max-count=1)` (which if you backport a fix to an older version like v0.2.1 may not actually be most current compared to v0.3.0!)  If someone is using an older commit but has fetched the latest remote git history, the appropriate semver is the preceding one to the current commit, but if it's a few commits after (like this one)... then is the version `v0.3.0-39ece4c`?
You can find the current git SHA1 using  `git rev-parse HEAD`.

Our current tooling and documentation also doesn't exactly encourage people to build a genqlient binary where a version would easily be injected. 

You **can** do `go run -ldflags="-X generate.version v0.2.0" github.com/Khan/genqlient` to inject it inside of a go generate directive, but that could then be a lie if they updated the project's `go.mod` and failed to update their go generate directive to match.

It seems like we can merge this (once the tests succeed) and iron out further kinks later.

generate/config.go
csilvers: To my mind, the correct Version is whatever would end up in go.mod if you did `go get genqlient@<head>`.  I don't know how easy that is to calculate though.

StevenACoffman:
So *outside* of genqlient itself (like in an application that uses genqlient as a library and as a tool) I think you could do this:
```
//go:generate go run github.com/Khan/genqlient -ldflags="-X generate.version $(go list -m -f '{{ .Version }}' github.com/Khan/genqlient)" 
```
Inside of genqlient, as in when you are building a binary, the main module contains no version information for that command, so you would need something like:
```
TZ=UTC git --no-pager show \
>   --quiet \
>   --abbrev=12 \
>   --date='format-local:%Y%m%d%H%M%S' \
>   --format="%cd-%h"
```

But I think it would be better to call out to the Go tools somehow to get it.

StevenACoffman:
Ah ha! 

The thing with [pseudo-versions](https://golang.org/ref/mod#pseudo-versions) is that it's just an encoding for a particular revision in a VCS (not just Git).

> Pseudo-versions may refer to revisions for which no semantic version tags are available. They may be used to test commits before creating version tags, for example, on a development branch.

So to have a pseudo-version, to begin with you must refer to **a revision**. The tip of the `master` branch might good enough, therefore:

```
go list -f '{{.Version}}' -m example.com/mymodule@master
```

Notice the syntax `module@version`, which allows you to choose a specific revision, hence, get the corresponding pseudo-version.

So to build the binary with the injected version like @csilvers suggested:
```
go build -ldflags="-X generate.version $(go list -f '{{.Version}}' -m github.com/Khan/genqlient@$(git rev-parse HEAD))" -o mytool ./main.go
```
Or:
```
go build -ldflags="-X generate.version $(go list -f '{{.Version}}' -m github.com/Khan/genqlient@main)" -o mytool ./main.go
```

StevenACoffman:
See https://github.com/Khan/genqlient/pull/129 where I continued our adventures by updating the expectations and Makefile according to my comments above.

generate/config.go
StevenACoffman: Inside genqlient it is:
```
go list -f '{{.Version}}' -m github.com/Khan/genqlient@$(git rev-parse HEAD)
```
Outside genqlient in an application that has genqlient in it's go.mod, it is:
```
go list -m -f '{{ .Version }}' github.com/Khan/genqlient)" 
```

benjaminjkraft:
I think the discussion here is getting too complicated.  I looked at what a few other libraries do and:
- [oapi-codegen -v](https://github.com/deepmap/oapi-codegen/blob/master/cmd/oapi-codegen/oapi-codegen.go#L85) and [in its "code generated by" line](https://github.com/deepmap/oapi-codegen/blob/f979a85e2c4333fc1fb96262929b5768ad1b8098/pkg/codegen/codegen.go#L546) does exactly what this PR does
- [golangci-lint version](https://github.com/golangci/golangci-lint/blob/master/cmd/golangci-lint/mod_version.go) does something similar
- [gqlgen version](https://github.com/99designs/gqlgen/blob/5ad012e3d7be1127706b9c8a3da0378df3a98ec1/graphql/version.go) uses a hardcoded version
I think we should take one of those approaches; I lean towards the former but I can see advantages both ways.  (Optionally we could set it up in such a way that you can override it via ldflags; it's unclear to me how well either approach handles untagged versions.  But I don't think that should be the primary mechanism for release versions.)

The only problem with this approach as it stands is the tests, which apparently somehow get `(devel)` locally and `<unknown>` in github actions.  I think the best approach there is to just add a (possibly undocumented) option to omit the version, and use that in most of our tests, which don't really benefit from this behavior.  I can take a look at doing that sometime (but maybe not today).

benjaminjkraft:
1.18 adds simpler ways to do this!

PR #129, opened 2021-10-03
Redo makefile to inject version, fix snapshots, update docs

StevenACoffman:
I broke this out of https://github.com/Khan/genqlient/pull/127 so this is my first stacked diff.
- Update snapshots to make tests pass
- Redo Makefile and inject version into building/installing
- Update documentation for injecting version into `go run`-ing
- Add makefile niceties like `make help` and advice from https://tech.davis-hansson.com/p/make/

This presumes that we want to continue to maintain a Makefile. If we want to use [Mage](https://magefile.org/) I would be more than happy to do that instead, as Mage would probably work better on a windows machine that didn't have WSL installed.


csilvers:
I'll wait to see if ben wants to comment on this -- in general I'm happy to keep ben as the main maintainer if he wants to be!  I'm not sure what thoughts he had.

I will point out that this doesn't do anything for our use at Khan, which runs genqlient using `go run`.  In general, I wonder if we should just go with the simpler solution of hard-coding the version number in the library, and using that.  It's probably already hard-coded somewhere.

StevenACoffman:
@benjaminjkraft What are your thoughts?

dnerdy (changes requested):
(Also requesting changes to get this PR off my queue.)

benjaminjkraft:
1.18 adds simpler ways to do this!

PR #133, opened 2021-10-04
Allow creating aliases for builtin types, using `typename`.

### LINKED ISSUE #130: `typename` doesn't work on basic types.
It seems like the `typename` directive only works on structs.  But it would be great if it worked on basic types like `string`, so you could do:
```
query x {
    # @genqlient(typename: "RenderType")
    renderType
}
```
where `renderType` is a string in the graphql schema, and the generated code would be:
```
type RenderType string
struct x {
    renderType RenderType
}
```

csilvers:
## Summary:
This lets you write code like:
```
query x {
   # @genqlient(typename: "MyString")
   someStringField
}
```
and genqlient will do
```
typename MyString string
type x struct {
   someStringField MyString
}
```

This was not difficult to implement, though it required introducing a
new identifier type.  The main difficulty I had was weird test
failures, that it turns out was due to the tests putting a bunch of
fields on the same line, so that the genqlient directive on the
previous line applied to all of them, accidentally.  This became a
problem when `typename` suddenly started being respected for builtin
types!  I fixed it by just spreading out the queries a bit.

Fixes #130

## Test plan:
make check

### DIFF
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -481,13 +481,13 @@ func (g *generator) convertDefinition(
 	case ast.Scalar:
 		if builtinTypes[def.Name] != "" {
 			// In this case, the user asked for a custom Go type-name
-			// for a built-in type, e.g. `typename MyString string`.
+			// for a built-in type, e.g. `type MyString string`.
 			goType := &goTypenameForBuiltinType{
-				GoTypename:    name,
+				GoTypeName:    name,
 				GoBuiltinName: builtinTypes[def.Name],
 				GraphQLName:   def.Name,
 			}
-			return g.addType(goType, goType.GoTypename, pos)
+			return g.addType(goType, goType.GoTypeName, pos)
 		}
 
 		// (If you had an entry in bindings, we would have returned it above.)
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -289,7 +289,7 @@ func (g *generator) convertDefinition(
 		}, err
 	}
 	goBuiltinName, ok := builtinTypes[def.Name]
-	if ok {
+	if ok && options.TypeName == "" {
 		return &goOpaqueType{GoRef: goBuiltinName, GraphQLName: def.Name}, nil
 	}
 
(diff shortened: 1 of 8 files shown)


dnerdy:
@csilvers did you try using `bind` to map builtin scalars to nonstandard types? I believe `bind` is meant to address this case (whereas typename is meant to override the generated name for a type). See https://github.com/Khan/genqlient/blob/59b6df6aabdf79cca684dd709fa455b5489352ff/docs/genqlient.yaml#L79-L83 and https://github.com/Khan/genqlient/blob/main/docs/genqlient_directive.graphql#L152-L171.

csilvers:
The difference is that `bind` requires you to define the type yourself, which can be inconvenient. (In this car we need a separate package just to define the type name.) `typename` lets genqlient define it in the genqlient package.

generate/convert.go
```diff
@@ -479,6 +479,17 @@ func (g *generator) convertDefinition(
 		return g.addType(goType, goType.GoName, pos)
 
 	case ast.Scalar:
+		if builtinTypes[def.Name] != "" {
+			// In this case, the user asked for a custom Go type-name
+			// for a built-in type, e.g. `typename MyString string`.
```
dnerdy, line 484:
```suggestion
			// for a built-in type, e.g. `type MyString string`.
```

generate/convert.go
```diff
@@ -479,6 +479,17 @@ func (g *generator) convertDefinition(
...
+			goType := &goTypenameForBuiltinType{
+				GoTypename:    name,
```
dnerdy, line 486: nit: I think `GoTypeName` would be more consistent with the rest of the code.

generate/types.go
```diff
@@ -65,6 +65,14 @@ type (
 		GraphQLName            string
 		Marshaler, Unmarshaler string
 	}
+	// goTypenameForBuiltinType represents a builtin type that was
+	// given a different name due to a `typename` directive.  We
+	// create a type like `type MyString string` for it.
+	goTypenameForBuiltinType struct {
```
dnerdy, line 71: I might also name this goTypeForBuiltinType.

dnerdy (commented):
Got it. Okay, this looks reasonable for that case. Could you add an entry to the change log? https://github.com/Khan/genqlient/blob/59b6df6aabdf79cca684dd709fa455b5489352ff/docs/CHANGELOG.md#L25

csilvers:
I'm still not sure this is a good idea.  It seems less than ideal to have two ways to override basic types, one using `bind` and one using `typename`.  In general Ben is not a big fan of `bind`, because it turns off all the type-checking that genqlient normally does, so I am leaning toward this being a better approach.  I'll change the documentation to be clearer that `typename` and `bind` can both be used to replace built-in types, but `typename` is preferred.

csilvers:
Documentation added -- if it looks ok can you approve?

StevenACoffman (approved):
LGTM now!

generate/types.go
benjaminjkraft: (I would have called it `goNamedType`, but any of these work!)

PR #134, opened 2021-10-04
⚡️ Accept array of string as valid input schema path.

### LINKED ISSUE #88: Feature Request: Support multiple schema files
While consistency with gqlgen is really a goal as such, but it would be convenient if we similarly supported multiple schema files they way it does:
```
package: genqlient
schema:
- ../../pkg/graphql/shared-schemas/side_by_side_directives.graphql
- ../../pkg/graphql/shared-schemas/user_deletion.graphql
- ../../pkg/graphql/shared-schemas/automap.graphql
- schema.graphql
- homepage_render.graphql
# The files from which we pull operations (relative to genqlient.yaml).
operations:
- genqlient.graphql
```
A client that is generated from the federated `composed_schema.graphql` has a lot more churn than what is required to communicate with a single service.
It might be nice to generate a "just this service" client, but that would require supporting multiple schema files.

hsblhsn:
This PR adds a feature that accepts both string and array of string as valid input in `schema` key in config.

This feature is completely backward compatible. It does not deprecate or break the previous config file format.

Most of the code is sourced from [https://github.com/99designs/gqlgen](https://github.com/99designs/gqlgen). 

Closes #88

### DIFF
--- a/generate/config.go
+++ b/generate/config.go
@@ -1,7 +1,6 @@
 package generate
 
 import (
-	"fmt"
 	"go/token"
 	"io"
 	"io/ioutil"
@@ -166,13 +165,13 @@ func loadSchemaSources(schemas StringList) ([]*ast.Source, error) {
 
 				return nil
 			}); err != nil {
-				return nil, fmt.Errorf("failed to walk schema at root %s: %w", pathParts[0], err)
+				return nil, errorf(nil, "failed to walk schema at root %s: %w", pathParts[0], err)
 			}
 		} else {
 			var err error
 			matches, err = filepath.Glob(f)
 			if err != nil {
-				return nil, fmt.Errorf("failed to glob schema filename %s: %w", f, err)
+				return nil, errorf(nil, "failed to glob schema filename %s: %w", f, err)
 			}
 		}
 
@@ -189,7 +188,7 @@ func loadSchemaSources(schemas StringList) ([]*ast.Source, error) {
 		var schemaRaw []byte
 		schemaRaw, err = ioutil.ReadFile(filename)
 		if err != nil {
-			return nil, fmt.Errorf("unable to open schema: %w", err)
+			return nil, errorf(nil, "unable to open schema: %w", err)
 		}
 
 		source = append(source, &ast.Source{Name: filename, Input: string(schemaRaw)})
--- a/generate/config.go
+++ b/generate/config.go
@@ -1,12 +1,16 @@
 package generate
 
 import (
+	"fmt"
 	"go/token"
 	"io"
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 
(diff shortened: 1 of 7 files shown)


generate/stringlist.go
```diff
@@ -0,0 +1,33 @@
+package generate
+
+// StringList provides yaml unmarshaler to accept both `string` and `[]string` as a valid type.
```
csilvers, line 3: Nice!

generate/config.go
```diff
@@ -122,3 +127,72 @@ func initConfig(filename string) error {
...
+			schemas = append(schemas, m)
+		}
+	}
+	for _, filename := range schemas {
+		filename = filepath.ToSlash(filename)
+		var err error
+		var schemaRaw []byte
+		schemaRaw, err = ioutil.ReadFile(filename)
+		if err != nil {
+			return nil, fmt.Errorf("unable to open schema: %w", err)
```
csilvers, line 192: I think genqlient has its own `errorf` wrapper -- is it appropriate to use that here?

csilvers (changes requested):
Thanks, this look good to me!

The only thing that's needed before landing is to update docs/CHANGELOG.md, to add this new feature under `### New features:`.  There's also probably some documentation elsewhere in docs/ that needs to be updated to document the new yaml file format possibility.

generate/config.go
hsblhsn: > is it appropriate to use that here?


You are right. Whole code should have one unified style. Pushed f471c95.

hsblhsn:
How does this sound to you?


```
# The filename with the GraphQL schema (in SDL format), relative to
# genqlient.yaml. This can also be a list of filenames.
# such as:
#  schema:
#  - user.graphql
#  - ./schema/*.graphql
#  - ./another_directory/**/*.graphqls
schema: schema.graphql
```

It would be more specific. WDYT?

csilvers:
> How does this sound to you?
> 
> ```
> # The filename with the GraphQL schema (in SDL format), relative to
> # genqlient.yaml. This can also be a list of filenames.
> # such as:
> #  schema:
> #  - user.graphql
> #  - ./schema/*.graphql
> #  - ./another_directory/**/*.graphqls
> schema: schema.graphql
> ```
> 
> It would be more specific. WDYT?

That looks great!  It makes clear you can use glob patterns, too.

csilvers (approved):
Awesome!  Thanks so much.

I don't know if I actually have the ability to merge this in, but hopefully someone else on this review does if I don't.

dnerdy:
Hello @hsblhsn! Thanks for the contribution. Could you sign the [Khan Academy Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSdyXYrc8ogVoA46J9KXyIj5nKlZzNkOnQG-4A1R7X_BWGTShQ/viewform) when you have a moment?

PR #137, opened 2021-10-05
Align support for multiple operation-files and multiple schema-files

benjaminjkraft:
Multiple schema-files are now supported as of #134, but the support was
a bit different from how we did multiple operation-files.  Before anyone
starts to depend on the ways the syntaxes differ, let's just make them
the same.  Since it's easy, I also added support for having just a
single operations-file.

I also realized while writing this that the type-change is technically
breaking (if you call from Go), so documented it as such. I think
this is unlikely to affect many people.

Test plan: make check

### DIFF
--- a/generate/config.go
+++ b/generate/config.go
@@ -6,10 +6,7 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
-	"regexp"
-	"strings"
 
-	"github.com/vektah/gqlparser/v2/ast"
 	"gopkg.in/yaml.v2"
 )
 
@@ -21,7 +18,7 @@ type Config struct {
 	// The following fields are documented at:
 	// https://github.com/Khan/genqlient/blob/main/docs/genqlient.yaml
 	Schema           StringList              `yaml:"schema"`
-	Operations       []string                `yaml:"operations"`
+	Operations       StringList              `yaml:"operations"`
 	Generated        string                  `yaml:"generated"`
 	Package          string                  `yaml:"package"`
 	ExportOperations string                  `yaml:"export_operations"`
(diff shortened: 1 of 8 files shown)


csilvers (commented):
It looks like you don't support `**` anymore? (Or does Glob do it by default these days?) That's fine, I think, but you'll probably want to update the doc example that has one.

generate/config.go
```diff
@@ -126,72 +123,3 @@ func initConfig(filename string) error {
...
-		} else {
-			var err error
-			matches, err = filepath.Glob(f)
-			if err != nil {
-				return nil, errorf(nil, "failed to glob schema filename %s: %w", f, err)
-			}
-		}
-
-		for _, m := range matches {
-			if schemas.Has(m) {
```
csilvers, line 179: You can probably get rid of the `Has` function too.

csilvers (approved):
Great!  You may want to cc the original patch author so they know of the semantic change around `**`.

benjaminjkraft:
cc @hsblhsn if you wanted to take a look!  If it's an issue we could move to using the fancier globbing for both.

generate/parse.go
```diff
@@ -16,20 +16,30 @@ import (
...
+			return nil, errorf(nil, "unreadable schema file %v: %v", filename, err)
+		}
+		sources[i] = &ast.Source{Name: filename, Input: string(text)}
+	}
+
 	schema, graphqlError := gqlparser.LoadSchema(sources...)
 	if graphqlError != nil {
-		filename, _ := graphqlError.Extensions["file"].(string)
-		return nil, errorf(nil, "invalid schema file %v: %v", filename, graphqlError)
+		return nil, errorf(nil, "invalid schema: %v", graphqlError)
```
hsblhsn, line 36: Removing file name from error message will make it really hard to find issues in schema or operation files. Since we still do not have strong IDE support for graphql SDL files.

WDYT?

generate/testdata/snapshots/TestGenerateErrors-InvalidSchema-go
```diff
@@ -1 +1 @@
-testdata/errors/InvalidSchema.schema.graphql:4: invalid schema file testdata/errors/InvalidSchema.schema.graphql: Expected :, found }
+testdata/errors/InvalidSchema.schema.graphql:4: invalid schema: Expected :, found }
```
hsblhsn, line 1: Finding the exact issue will be tough just by seeing the new error message. Previous message was better IMO.

hsblhsn (changes requested):
Just my thoughts. Don't make it hard to debug.

generate/testdata/snapshots/TestGenerateErrors-InvalidSchema-go
csilvers: I think the filename is still there, just in a different place:
```
 testdata/errors/InvalidSchema.schema.graphql:4: invalid schema: Expected :, found } 
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

hsblhsn (approved):

PR #138, opened 2021-10-12
Allow the use of `typename` to be combined with `bind`.

csilvers:
## Summary:
In that case, we make a type-alias that points to the `bind` type, and
use that type-alias for the type of the field.

## Test plan:
make check

### DIFF
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -229,6 +229,21 @@ func (g *generator) convertType(
 		goRef, err := g.ref(localBinding)
 		// TODO(benkraft): Add syntax to specify a custom (un)marshaler, if
 		// it proves useful.
+		if options.TypeName != "" {
+			// You can combine bind and typename; we make (and use) a
+			// type-alias to the bound type instead of the bound type
+			// directly.
+			goType := &goTypenameForBuiltinType{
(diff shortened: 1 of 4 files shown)


csilvers:
@benjaminjkraft here's the diff, as suggested in #133.  There's a bit of code duplication, but I couldn't tell if it was worth trying to extract it out into a shared function, it seemed like it wouldn't actually help that much.  I'm wondering if it would make sense to try to combine the local vs global `bind` parsing to make that cleaner, but I didn't try that here.

I couldn't figure out if there were existing tests of the global bind option that I could hook into, to test the change for that.  If there are I didn't see them!

csilvers:
> I also realize that even moreso than I thought, this will be a bit confusing because named types don't inherit the JSON-marshaling. Maybe it's better to just reject the combination, since it's probably not what you intended. If people find use cases then we can reintroduce it.

Ooh yeah, that's a good point about the marshaling.  I think you're right it's better to just reject it for now, it doesn't seem very useful.  I'll do that when I have a chance.

csilvers:
>  I think you're right it's better to just reject it for now, it doesn't seem very useful.

Closing this as a result.  We should probably detect and reject this situation, but that can be a separate PR.

PR #154, opened 2021-11-05
TYPO: Fixed variable name and  position in interface definition.

### LINKED ISSUE #153: Few question about the client interface.
Here, where you have defined the interface, 

https://github.com/Khan/genqlient/blob/e0accbded177db9143314911bacedf61b2cda656/graphql/client.go#L41-L46

On line 45, we see `input, retval interface{}`. If I am not wrong, here `input` means the variables of the query and `retval` means the reference to the output struct.

So, the **first interface is input, and second one is output**.

But here in the implementation,

https://github.com/Khan/genqlient/blob/e0accbded177db9143314911bacedf61b2cda656/graphql/client.go#L91

We see the `retval, variables interface{}`, if i am not wrong the `retval` means output, and `variables` means input.

So, **the first interface is output, second interface is input**.

# The problem

On interface definition, the first var is indicating input. But in the implementation, the first var is indicating output.

I am confused.

hsblhsn:
I was also confused about the variable names (#153). But in this [commit](https://github.com/Khan/genqlient/commit/5995653583af6e8d42477652f7d51ac4201b9fe6) it is clear that the first var is output, second var is input. In this commit I have fixed the issue. Fixes #153

### DIFF
--- a/graphql/client.go
+++ b/graphql/client.go
@@ -42,7 +42,7 @@ type Client interface {
 		ctx context.Context,
 		opName string,
 		query string,
-		input, retval interface{},
+		retval, input interface{},
 	) error
 }
 


csilvers (approved):
Seems right to me!

Issue #149, opened 2021-10-28
pointer & omitempty is cumbersome/impossible with a large schema


nathanstitt:
First let me thank you for your great work on genqlient, it has a LOT of promise and fills a big need in the golang GraphQL space.  I've been searching for something like it for quite awhile.

In my opinion, this behavior would be ideal:
  * all references to other relationships should default to a pointer 
  * omitempty be allowed even if the arguments are required.

I've added an example repo that may be easier to follow than the below verbiage.  It's at https://github.com/nathanstitt/genqlient-stress-test

My reasoning for this is:   

I have a project that's using the Hasura Graphql Engine https://hasura.io which generates a schema for query/mutations on database tables.  Hasura has a very comprehensive query language that supports nesting and/or type queries using GraphQL types.

I've been using github.com/machinebox/graphql with the project for some time but would love to have it be type safe and not require me to hand craft structs that match the existing GraphQL schema.  

I attempted to migrate to genqlient but ran into several issues that you may find enlightening if I explain them.

My schema has nested types like so:
```GraphQL
input client_bool_exp {
  _and: [client_bool_exp!]
  _not: client_bool_exp
  _or: [client_bool_exp!]
}
```

This requires `_and`, `_not`, and `_or` to be a pointer, otherwise the generated code won't compile due to:`invalid recursive type Client_bool_exp` errors.

I ended up with a list of `122` `@genqlient(for: "<struct>._not", pointer: true)` statements in order to get it to compile, and I only have 35 tables 😭 

Once it was compiling, I then encountered issues with mutations inserting undesired zero values into the database because `omitempty` wasn't set.  

Rather than attempt to repeat the above pointer exercise, I set all mutations to be prefixed with `@genqlient( omitempty: true)`. That failed because the inputs were required, and if I attempted to use `for`,  it then fails with: `for is only applicable to operations and arguments`.   I made a tiny PR to allow that https://github.com/Khan/genqlient/pull/148 but @csilvers suggested I first make an issue to discuss first.

FWIW: https://github.com/Khan/genqlient/issues/15 also touches on this subject as well

I'm attaching my 25K line schema just in case you want to review it.  If you really want to torture yourself, attempt to write this mutation against it:
```GraphQL
mutation insertAccount(
  $tempUserId: bigint!,
  $account: account_insert_input!
) {
  insert_account_one(object: $account) {
	id
  }
  delete_user_temp(where: { id: {_eq: $tempUserId} }) {
	affected_rows
  }
}
```

[...]

PR #155, opened 2021-11-08
Add "struct_references" configuration

nathanstitt:
Discussion around this: https://github.com/Khan/genqlient/issues/149  

Adds a "use_weak_references" configuration that will:

  * Use a pointer type for struct fields that are a complex type
    * The behaviour can be overridden by setting "pointer: false" in the
      proceeding comment blocks
  * Sets the "omitempty: true" flag on fields matching the criteria.
    This can also be overriden by setting "omitempty: false"

Comments & suggestions are very welcome, I'm not at all sure this is the best way to implement this.

Although we'd debated not setting the pointer for array elements, it did turn out to be simpler to set them everywhere and also made the documentation cleaner to have a single rule to explain.


docs/genqlient.yaml
```diff
@@ -77,6 +77,17 @@ context_type: context.Context
 # without making a query.
 client_getter: "github.com/you/yourpkg.GetClient"
 
+
+# if set, fields with a complex type will default to having
+# the "pointer: true, omitempty: true" flag
```
benjaminjkraft, line 82:
```suggestion
# the "pointer: true, omitempty: true" flag.
```

docs/genqlient.yaml
```diff
@@ -77,6 +77,17 @@ context_type: context.Context
 # without making a query.
 client_getter: "github.com/you/yourpkg.GetClient"
 
+
+# if set, fields with a complex type will default to having
```
benjaminjkraft, line 81:
```suggestion
# If set, fields with a struct type will default to having
```
(I think it's worth being explicit what types we mean.)

docs/genqlient.yaml
```diff
@@ -77,6 +77,17 @@ context_type: context.Context
...
+# the "pointer: true, omitempty: true" flag
+#
+# This can be useful for complex schema where it would be burdensome
+# to manually set the flags on a large number of fields
```
benjaminjkraft, line 85:
```suggestion
# to manually set the flags on a large number of fields.
```

generate/convert.go
```diff
@@ -235,16 +236,23 @@ func (g *generator) convertType(
...
 	def := g.schema.Types[typ.Name()]
+
 	goTyp, err := g.convertDefinition(
 		namePrefix, def, typ.Position, selectionSet, options, queryOptions)
 
-	if options.GetPointer() {
+	if g.getWeakReference(options, isStructField, def) {
+		if options.Omitempty == nil {
+			oe := true
+			options.Omitempty = &oe
```
benjaminjkraft, line 252: Mutating this seems a bit awkward.  Can we instead just put this in where we call `GetOmitempty` as well?  That will also help with the comment below I suspect.

generate/convert.go
```diff
@@ -253,6 +261,18 @@ func (g *generator) convertType(
 	return goTyp, err
 }
 
+// getWeakReference decides if a field should be of pointer type and have the omitempty flag set.
+func (g *generator) getWeakReference(
+	options *genqlientDirective,
+	isStructField bool,
+	def *ast.Definition,
+) bool {
+	return options.Pointer == nil &&
```
benjaminjkraft, line 270: I think this will be a little confusing in that it will mean if you do `omitempty: false` it will be ignored, but if you do `pointer: false` it will not only work but also turn off `omitempty`.  (And in fact if you do `pointer: true` it will also turn off omitempty!)  If we can, I think it'll be better to check each setting separately, so that if you set `pointer` explicitly that overrides `pointer`, like you have it, but not `omitempty`.

generate/convert.go
```diff
@@ -253,6 +261,18 @@ func (g *generator) convertType(
...
+		isStructField &&
+		g.Config.WeakReferences &&
+		len(def.Fields) > 0
```
benjaminjkraft, line 273: Oh so actually a question here is whether `use_weak_references` should apply to interface-typed fields.  Unless you see a strong reason, my feeling is not -- pointer-to-interface types are super awkward in Go.  In that case this should be `def.Kind == ast.Object || def.Kind == ast.InputObject` (or something like that).)

generate/convert.go
```diff
@@ -174,7 +174,7 @@ func (g *generator) convertArguments(
 		// names.go) and the selection-set (we use all the input type's fields,
 		// and so on recursively).  See also the `case ast.InputObject` in
 		// convertDefinition, below.
-		goTyp, err := g.convertType(nil, arg.Type, nil, options, queryOptions)
+		goTyp, err := g.convertType(nil, arg.Type, false, nil, options, queryOptions)
```
benjaminjkraft, line 177: Oh interesting, do you think we don't want to apply this to arguments?  I would think we would; while of course there it's easier to set it seems like a more consistent default would be better.  If you feel this is clearer I'm ok with that.  Either way probably worth an explicit mention in the docs.

benjaminjkraft (changes requested):
Thanks so much for doing a PR of this!  The way you did it it's not nearly as awkward as I feared it might be.  As you say, I think it's worth trying to make the rules as simple as possible, so several of my comments are on things I feel are a bit inconsistent.

The one other thing that jumps out is I don't like the name.  (I know, you asked for my opinion earlier and I was noncommital, sorry.)  I don't have any great ideas and I don't think we'll find something perfect, but "weak references" has too much of a specific meaning elsewhere (i.e. a reference that doesn't protect the target from GC).   Maybe `struct_references: true` or `struct_default: pointer_and_omitemty` or something?

generate/convert.go
nathanstitt: Agreed, but I played with a few different options here but didn't find anything better.  GetOmitempty is called in a few spots, and we'd have to pass the Config and field definition into it and then duplicating the logic of the `getStructReference` check.

The updated check I've came up with here is a bit cleaner IMO, so see what you think

generate/convert.go
nathanstitt: Yep, you're correct.  The logic is slightly flawed here as you describe.  I'll rework

generate/convert.go
nathanstitt: Great catch, I hadn't considered the interface-typed fields.  I don't see why we shouldn't use pointers for those, that way the rule doesn't have any gotchas

generate/convert.go
nathanstitt: Hm, yes Interesting. I hadn't really considered allowing arguments, but don't see why not.  Plus, allowing it will simplify the code by removing the need to pass around the `isStructField` argument.

nathanstitt:
@benjaminjkraft ready for another look whenever you're able.  Sorry for the delay, I reworked my project to use this to double check it's all ok.   It worked great, I still had to add approx 10 "omitempty" statements but that's quite manageable.

arjunyel:
@nathanstitt hey I just wanted to thank you so much for doing this PR, you're helping a lot of people :D

generate/convert.go
benjaminjkraft: They just tend to be awkward in that you have to explicitly dereference to call the interface's methods (i.e. `(*resp.F).M()`) and confusing in that the value inside the interface is often itself a pointer.  I think it's clean to explain either way -- if anything since the option is "struct_references" that would suggest it's just structs.  We could rename it and do interfaces but I feel that's only going to be annoying.

Another question that we had discussed in the issue is whether it applies to output objects.  I think "what is easiest to explain" is also a good criterion there, but it could go either way (if inputs, we'd call it "input_references" and it makes the interfaces question moot since GraphQL doesn't have interfaces in inputs).

generate/convert.go
benjaminjkraft: Yeah, fair enough!

benjaminjkraft (approved):
Neat, this looks great!  One question about whether we should do this for output types at all -- I'm ok either way as long as the name matches so as-is is fine unless you want to change it.  (Long term maybe we'd do both as options if the need arises.)  And it looks like a couple of lint errors crept in, but once you fix those it's good to go!

Ah and I almost forgot (our bot is down) -- could you sign our CLA at https://www.khanacademy.org/r/cla?

nathanstitt:
Thanks @benjaminjkraft!   

I think output types should be affected by the rule because:
  1) It's a bit better (IMO) to check for missing association by testing for nil vs making sure the pkey field is included and then checking it for it's zero value.     However, this is a bit unsafer and runs a risk of panics, so it really could go either way.  Personally I'd rather the code blow up if there's a misconception about a missing value vs continuing to run with garbage data.  I could probably be convinced either way though so 🤷 
  2) It's better to use a single rule everywhere if possible.  That's both easier to document and also to keep straight mentally.

nathanstitt:
Oh and I also signed CLA, so should be good to merge whenever you're good with it

benjaminjkraft:
Thanks very much!

PR #156, opened 2021-11-26
Sort operations to guarantee a stable order

edigaryev:
Problem: currently when using a wildcard that covers multiple files in `operations:` YAML directive the generator emits functions in a non-stable ordering due to the use of map for storing file names:

https://github.com/Khan/genqlient/blob/e0accbded177db9143314911bacedf61b2cda656/generate/parse.go#L68-L84

Solution: sort the operations before emitting them.


csilvers:
Good catch! Change looks good to me. The only issue before merging is to make sure you've signed the CLA. Sadly I don't know the details of that. I'll try to figure out out after the holiday (probably Monday)

csilvers:
Oh, one more thing: it looks like there's a failing test. Can you fix that up?

benjaminjkraft:
Good idea -- matches what we do for the types they reference.  Thanks!

The CLA is at https://www.khanacademy.org/r/cla.   For the tests I think you'll just need to `go generate ./...` to rebuild the example and integration tests.

edigaryev:
> Oh, one more thing: it looks like there's a failing test. Can you fix that up?

Fixed a couple of tests failing locally, see https://github.com/Khan/genqlient/pull/156/commits/b9c47e446f5c752f3271a350a0ed2639b88df570 and https://github.com/Khan/genqlient/pull/156/commits/95c5403cb3becb7333aaa43e14c89f436749dd42.

Not sure about the tests that run on CI, though.

edigaryev:
>For the tests I think you'll just need to go generate ./... to rebuild the example and integration tests.

See https://github.com/Khan/genqlient/pull/156/commits/69302258e58b61834cf607e895c00415feb7e752.

edigaryev:
> The CLA is at https://www.khanacademy.org/r/cla.

CLA sent.

PR #161, opened 2021-12-18
Move 'Code generated by' disclaimer up

adambabik:
`Code generated by` disclaimer should be at the top of the file, as it is demonstrated in [this article](https://go.dev/blog/generate) on The Go Blog.

The current placing breaks integration with some code formatters and linters which do no skip files generated by `genqlient`.


csilvers:
Thanks! I know it's a small change but can you sign the CLA of you haven't already?  https://www.khanacademy.org/r/cla

benjaminjkraft:
Ah, thanks for noticing! It looks like the rule changed with golang/go#41196, which is why I remember it being anywhere in the file but in fact the [go generate docs](https://pkg.go.dev/cmd/go#hdr-Generate_Go_files_by_processing_source) now clearly say
> This line must appear before the first non-comment, non-blank text in the file.

I've just clicked to let CI run; I think you'll need to update some of the text fixtures to match, which you can do via e.g.
```sh
go generate ./...
UPDATE_SNAPSHOTS=1 go test ./...
```
Once you've done that, looks great to me!

adambabik:
Thanks for the guidelines. I updated snapshots and rerun tests - they passed locally.

Issue #164, opened 2022-01-19
Interfaces that implement other interfaces


jmatichard:
Hello `genqlient` team !

We are trying to generate queries from a GraphQL schema that contains interfaces that implements other interfaces.

According to the GraphQL specs, this should be valid (cf. [issue #373](https://github.com/graphql/graphql-spec/pull/373) in the GQL repository).

Here is our (simplified) schema:

```
scalar DateTime

interface BaseEntity {
  id: String!
  createdAt: DateTime!
  updatedAt: DateTime!
}

interface Transaction implements BaseEntity {
  id: String!
  createdAt: DateTime!
  updatedAt: DateTime!
  userId: String!
  productId: String!
}
```

and the `genclient.yaml` config:

```
schema: schema.graphql
operations:
  - genqlient.graphql
generated: generated.go
package: main
bindings:
  DateTime:
    type: time.Time
```

When we the run the `go run github.com/Khan/genqlient` command, we get the following error message:

> schema.graphql:9: invalid schema file schema.graphql: Unexpected Name "implements"

Can you please confirm that `genqlient` doesn't currently support this feature ?

Do you have some hints so we can help implement this in the codebase ?

Thank you very much for your hard work on this Go library !

benjaminjkraft:
Thanks for reporting!  This is blocked by the GraphQL parser not supporting interfaces which implement interfaces (vektah/gqlparser#166).  In theory genqlient should be more or less ready to support this once gqlparser does, although it's hard to be sure without being able to test it!

StevenACoffman:
PRs to address vektah/gqlparser#166 are very welcome btw!

benjaminjkraft:
This is no longer blocked on gqlparser.  I have some work in progress to support it on the genqlient side, but I ran into a few snags.  I'll hopefully have more time to work on it soon!

stiray:
@benjaminjkraft, did the snags go away? :)

The thing is that I would love to use genqlient for next project which has just started but since schema is full of "implements", I would have to use something else.

cameronbrill:
@benjaminjkraft similar situation to stiray. starting a new project and would like to use this, but the schema I'm referencing is full of implements. any updates?

alfaindia:
I also have full of implements when trying to use linear graphql api

cameronbrill:
[...]

PR #166, opened 2022-01-26
gqlparser v2.3.1

StevenACoffman:
gqlparser v2.3.1 appears to work ok, but  v2.3.0 gqlparser  had a PR that needed to be reverted.

FYI I cut a new v2.3.0 gqlparser release https://github.com/vektah/gqlparser/ thinking it had been a really long time, but it looks like the https://github.com/vektah/gqlparser/pull/112 might have been half baked as it causes problems for gqlgen and genqlient:
```
go get -d github.com/vektah/gqlparser/v2@v2.3.0
go get github.com/99designs/gqlgen/cmd@v0.14.0
go mod tidy
go generate ./...
go test ./...
$ make example
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ require (
 	github.com/alexflint/go-arg v1.4.2
 	github.com/bradleyjkemp/cupaloy/v2 v2.6.0
 	github.com/stretchr/testify v1.7.0
-	github.com/vektah/gqlparser/v2 v2.2.0
+	github.com/vektah/gqlparser/v2 v2.3.1
 	golang.org/x/tools v0.1.7
 	gopkg.in/yaml.v2 v2.4.0
 )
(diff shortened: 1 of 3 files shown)


benjaminjkraft:
Thanks for testing and fixing the issue!  Is there a reason to upgrade in genqlient?  If there's a new feature we need or whatnot I'm happy to but I don't know if there's a reason to make our users upgrade if not (they can always just choose to upgrade in their own go.mod).

StevenACoffman:
Mostly vektah/gqlparser@v2.3.1 has bugfixes that end users may notice when using genqlient. I was looking at how hard it would be to implement #164 by fixing vektah/gqlparser#166 ( some initial work here: vektah/gqlparser#169 ) and I wanted to insure that future releases of vektah/gqlparser would not break genqlient.

## What's Changed in gqlparser v2.3.1 since v2.2.0
* make importer reproducible by @vvakame in https://github.com/vektah/gqlparser/pull/112
* Update prelude to match graphql_spec for @deprecated directive by @danielnovograd in https://github.com/vektah/gqlparser/pull/155
* Stop validator mutating the provided *Schema by @xzyfer in https://github.com/vektah/gqlparser/pull/158
* feat: do not mark __typename field as unknown in the schema validation by @damianopetrungaro in https://github.com/vektah/gqlparser/pull/172
* Fix: argument comparison between two sets of args by @jufemaiz in https://github.com/vektah/gqlparser/pull/170
* Fix for update by reference of default variable values by @aurelijusbanelis in https://github.com/vektah/gqlparser/pull/161

csilvers:
I feel ben should have the final word, but I'm fine with this fwiw.

benjaminjkraft:
Thanks for the explanation.  I thought about it more and you've convinced me -- since gqlparser is internal to genqlient it makes good sense for us to upgrade it even though users may also use it via other libraries (i.e. gqlgen).  If it doesn't work right they see that as our bug even if the cause is in gqlparser.

PR #170, opened 2022-02-07
Implement fragment dynamic

NuVivo314:
This proposal. 

The problem: Who make query dynamic return. 
Solution: The dynamic fragment.

### DIFF
--- a/example/generated.go
+++ b/example/generated.go
@@ -80,18 +80,20 @@ func getUser(
 	}
 	var err error
 
-	var retval getUserResponse
-	err = client.MakeRequest(
-		ctx,
-		"getUser",
-		`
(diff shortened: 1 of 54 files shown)


benjaminjkraft (changes requested):
Thanks for the PR! Again here I'd like to understand what it is you're trying to do -- in this case just reading the code didn't even give me a very good idea of what it does -- something about dynamically constructed queries? -- nor why you might want it.  Could you fill that in a bit more?  Perhaps both some more explanation of your use case, and/or some snapshot and integration tests -- we'll need those eventually anyway to merge.

NuVivo314:
Hello, i close this. Is very complex and don't add value.

PR #172, opened 2022-02-10
Reject the use of both typename and bind

benjaminjkraft:
In #133, Craig added support for a new use of typename, where it applies
to a scalar and means that genqlient should generate a named type, e.g.
`# @genqlient(typename: "MyString")` on a node of type string will
generate and use `type MyString string`.  But this gets a bit confusing
if you mix it with `bind`; should
`typename: "MyString", bind: "int32"` generate `type MyString int32`, or
should one override the other, or what?  Of course in practice you're
not likely to write that all in one place, but you could via a global
binding, or a `for` directive, and in that case probably it was a
mistake.  In #138, we looked at making them work together correctly, but
it added complexity and got even more confusing.

So instead, here, we just ban it; we can always add it back if it proves
useful.  (Or, you can make the `typename` win over a global binding by
locally unbinding it via `bind: "-"`.)  This required changes in
surprisingly many places; I already knew the directive-validation code
was due for a refactor but that will happen some other day.  The tests
show that it works, in any case.

Interestingly, this problem actually could have arisen for a struct
binding already, before #133.  But all the same reasons it's confusing
seem to apply, so I just banned it there too.  This is technically a
breaking change although I doubt anyone will hit it.

Test plan: make check

I have:
- [x] Signed the [Khan Academy CLA](https://www.khanacademy.org/r/cla) (hehe I realized I actually should go and sign this now!)
- [x] Included a link to the issue fixed, if applicable (n/a)

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -23,6 +23,7 @@ When releasing a new version:
 ### Breaking changes:
 
 - The `Config` fields `Schema` and `Operations` are now both of type `StringList`.  This does not affect configuration via `genqlient.yaml`, only via the Go API.
+- The `typename` and `bind` options may no longer be combined; doing so will now result in an error.  In practice, any such use was likely in error (and the rules for which would win were confusing and undocumented).
 
 ### New features:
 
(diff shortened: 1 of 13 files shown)


csilvers (approved):
Yay!  I'm so glad you did this instead of me, I'm sure I would have made a hash of it.  All the changes look good to me, though if there was another place you needed to validate but missed, I'm unlikely to have noticed.

benjaminjkraft:
Yeah, I'm only 90% sure I thought of all the ways it could happen, but I decided that's way better than 0%!

PR #173, opened 2022-02-10
Reject a glob that matches no files

### LINKED ISSUE #146: On main, there is no error if the schema file cannot be found
**Describe the bug**
In v0.3.0 if you point to a schema file that does not exist you will be given an error:
```
unreadable schema file graph/schema.graphql: open graph/schema.graphql: no such file or directory
```

Past v0.3.0, you get a confusing error that the types are not known: 
```
graph/genqlient.graphql:35: query-spec does not match schema: Unknown type "ActionInput".
```

**To Reproduce**
1. `genqlient.yaml` contains a schema path that does not exist: `schema: graph/nonexistent.graphqls`
2. Create an operations file that references types defined in the schema
3. Run `go run github.com/Khan/genqlient`

**Expected behavior**
```
unreadable schema file graph/schema.graphql: open graph/schema.graphql: no such file or directory
```

**genqlient version**
`v0.3.1-0.20211022224026-a4aa6d9bb0f4`

The problem cannot be reproduced on `v0.3.0`.

**Additional context**
I believe the behavior was introduced in this commit: 59b6df6aabdf79cca684dd709fa455b5489352ff

Since we are globbing, we end up with an empty list of files to iterate. Since we don't try to load any files, we never return an error that they can't be loaded.

I'm not sure the best fix here - perhaps try to detect if a line is a glob or not, and only call the glob function if it is? Or maybe a glob pattern that returns no results should result in an error - or warning.

benjaminjkraft:
The most important case here is if your glob isn't even a glob, it's
just a filename.  But even if it was a glob, it's probably a mistake;
you'll probably end up with a confusing error due to an empty schema, or
a slightly less confusing error due to not having any operations.
Instead, let's just say outright that your glob didn't match any files.

Fixes #146.

Test plan:
This was a bit annoying to test via snapshot, so I just tested it
manually by modifying the example to use a glob that didn't match any
files, first for the schema then for the operations, and got errors like
```
bogus*.graphql did not match any files
exit status 1
example/main.go:68: running "go": exit status 1
```

I have:
- [x] Added tests covering my changes, if applicable (see above)
- [x] Included documentation, for new features (n/a)
- [x] Added an entry to the changelog (n/a, fixing a bug not in a release)

### DIFF
--- a/generate/parse.go
+++ b/generate/parse.go
@@ -72,6 +72,9 @@ func expandFilenames(globs []string) ([]string, error) {
 		if err != nil {
 			return nil, errorf(nil, "can't expand file-glob %v: %v", glob, err)
 		}
+		if len(matches) == 0 {
+			return nil, errorf(nil, "%v did not match any files", glob)
+		}
 		for _, match := range matches {
 			uniqFilenames[match] = true
 		}


StevenACoffman (approved):
Hmmm, yeah tricky to test, so fine to just do it without coverage.

csilvers (approved):
Seems good to me!

PR #179, opened 2022-03-22
Update tested Go versions (and deps)

benjaminjkraft:
Go 1.18 is out! So we should run tests on it.

Additionally, gqlgen had some issues with it (see 99designs/gqlgen#1961
and golang/go#45584) so I updated that too, which updated some other
things.

Finally, latest gqlgen requires 1.16+, and it's time for us to do the
same anyway, so we can use `embed` and other newer goodies.  So I
dropped running tests for 1.14 and 1.15, and bumped the module language
version.

Test plan: make check

I have:


benjaminjkraft:
> Are you _sure_ you've signed the khan academy CLA? :-P

I did! I realized while writing the checklist that now I actually ought to!

> Change looks good to me. You may want to add the special `.gitattributes` thing that tells github to hide diffs for go.sum.

Ah, good idea.  I marked it generated so you can still expand the diff.

Also note I added to this removal of 1.14 and 1.15 (PR message now updated).

benjaminjkraft:
Yeah.  Makes sense for us too, so we can use embed (PR to follow).

PR #181, opened 2022-03-22
Remove ioutil

benjaminjkraft:
In 1.16, it's now deprecated and replaced by `io` and `os`.  Let's
upgrade!

Test plan:
- make check
- git grep ioutil

I have:
- [x] Written a clear PR title and description (above)
- [x] Signed the [Khan Academy CLA](https://www.khanacademy.org/r/cla)
- [x] Added tests covering my changes, if applicable
- [x] Included a link to the issue fixed, if applicable
- [x] Included documentation, for new features
- [x] Added an entry to the changelog

PR #182, opened 2022-03-22
Upgrade golangci-lint

benjaminjkraft:
1.42 seems to be having some problems with Go 1.18 (see #2649), so let's
just upgrade to latest.  There was one new error, suggesting to use `%q`
rather than `"%s"`, since the former does escapes properly.  It doesn't
really matter for us -- the string should be an identifier -- but it
doesn't hurt.  (Plus it wanted us to upgrade from ioutil, which I did
in #181.)

Test plan: make check

I have:


StevenACoffman (approved):

csilvers (approved):
Heh, I ran into that same problem trying to upgrade for webapp!  And issues with the `%q` linter (which I turned off for us, but may make sense for genqlient).

Issue #19, opened 2021-04-22
Simplify graphql.Client interface (MakeRequest)


benjaminjkraft:
The `graphql.Client` interface is scarily complex, for something we have to support forever.  We may want to try to make it a bit more constrained, with specific hooks for specific things, so that callers can't do weird things.  But we'll need to see how people actually use it.

At Khan, we currently:
- munge the URL to add the opname
- munge the error (if any), wrapping, and potentially adding some of the variables

Other potential use cases:
- caching (so #29 can be done in third-party code)
- send a hash (likewise #20) or minify (#7)
- non-HTTP transports
- non-JSON transports (this will be fairly hard since we'll have to generate unmarshalers for them; I'm not sure if anyone even does this though)

benjaminjkraft:
Another option is to do
```go
type Request struct {
  OperationName string
  Query string
  Variables map[string]interface{}
}
type Client interface {
  MakeRequest(ctx context.Context, request *Request, retval interface{})
}
```
and then we could at least add fields.

We could even replace `retval interface{}` with returning the `json.RawMessage`, or with passing you `unmarshal(string) interface{}` and having you return the `interface{}`, although both are a bit ugly and I'm not sure either is quite flexible enough.

benjaminjkraft:
A great solution similar to my above comment fell out in #184, so I think we are all set here.

Issue #183, opened 2022-03-23
Add support for extensions


janboll:
**Is your feature request related to a problem? Please describe.**
I want to access the data returned in the Extensions field, as defined in the response format: https://spec.graphql.org/October2021/#sec-Response-Format
 A challenge is, that this data is untyped and not defined in the schema itself.

**Describe the solution you'd like**
Add a configuration flag to `genqlient.yaml`, like `use_extensions`, to toggle generating Extensions fields in resulting objects.  Example for the resulting, generated code:
```
type GithubOrgsResponse struct {
	Githuborg_v1 []GithubOrgsGithuborg_v1GithubOrg_v1 `json:"githuborg_v1"`
	Extensions   map[string]interface{}
}
```
Extensions would be added after unmarshaling the data by the graphlql client. 

**Describe alternatives you've considered**
Enabling it on a per-query basis via comment 

**Additional context**
I am willing to provide a PR once the design is settled.

benjaminjkraft:
Thanks for raising this!  I think adding an option makes a lot of sense.  The only thing I want to discuss a little is where and how to represent it.

For where, in GraphQL generally the `extensions` should be at the same level as the `data` and `errors`.  So that would suggest to make it a separate result parameter parameter, i.e. we'd generate `GithubOrgsResponse` as-is but return `(resp GithubOrgsResponse, ext map[string]interface{}, err error)` or similar.  It's a bit unwieldy to have 3 results, but it also means we can avoid worrying about confusion (or even name-collision) between whether `Extensions` is a query field with that name, or actually extensions.  Does that seem sensible to you?  We could also add another layer of struct, but that seems awkward in its own way.

(One slight awkwardness there is that we may need to change the `Client` interface, again. In principle shoving `Extensions` into the generated `MyQueryResponse` avoids that, but I think in practice it will make the life of a `Client` harder, so a simple breaking change is probably actually less bad. Or we could have two options, `Client` and `ClientWithExtensions`? Possibly the best thing is actually to make `payload` and `response` public and then do `MakeRequest(ctx context.Context, req *payload, resp *response)` as a more extensible interface. cc @csilvers or @dnerdy can you take a quick glance at the Khan client to see if that would in fact work and be reasonably extensible? Anyway, I'm not too worried about this, I think we can work something out when we get there, possibly in a separate PR in the same release. )

[...]

PR #184, opened 2022-03-29
Adding GraphQL Extensions

janboll:
This enables accessing the Extensions field, as defined in the response
format: https://spec.graphql.org/October2021/#sec-Response-Format

Extensions can be enabled using configuration option use_extensions.
This will change the return parameters of the generated client
functions. Making it a breaking change if enabled.

Since extensions are untyped as defined in the spec, the Client will
return an interface of type map[string]interface{}.

Issue: https://github.com/Khan/genqlient/issues/183

I have:
- [X] Written a clear PR title and description (above)
- [X] Signed the [Khan Academy CLA](https://www.khanacademy.org/r/cla)
- [X] Added tests covering my changes, if applicable
- [X] Included a link to the issue fixed, if applicable
- [X] Included documentation, for new features


generate/operation.go.tmpl
```diff
@@ -12,34 +12,41 @@ func {{.Name}}(
...
-        "{{.Name}}",
-        `{{.Body}}`,
-        &retval,
-        {{if .Input}}&__input{{else}}nil{{end}},
+        req,
+        resp,
     )
-    return &retval, err
+
+	retval := resp.Data.(*{{.ResponseName}})
```
benjaminjkraft, line 50: Maybe better to keep a reference to this above, so you don't have to cast it?  i.e. above do:
```go
var data {{.ResponseName}}
resp := &graphql.Response{Data: data}
```
then here just `return data, ...`.

generate/operation.go.tmpl
```diff
@@ -12,34 +12,41 @@ func {{.Name}}(
...
     {{end -}}
     {{end -}}
-) (*{{.ResponseName}}, error) {
-    {{- if .Input -}}
+) (*{{.ResponseName}}, {{if .Config.Extensions -}}map[string]interface{},{{end}} error) {
+    req := &graphql.Payload{
+        OpName: "{{.Name}}",
+        Query:  `{{.Body}}`,
+    {{if .Input -}}
     {{/* We need to avoid conflicting with any of the function's argument names
```
benjaminjkraft, line 20: This comment no longer applies, you can just remove it!

generate/testdata/queries/foo.go
```diff
@@ -0,0 +1,69 @@
+// Code generated by github.com/Khan/genqlient, DO NOT EDIT.
```
benjaminjkraft, line 1: Is this used? Maybe left over from your testing?

generate/testdata/snapshots/TestGenerateWithConfig-ClientGetterNoContext#01-testdata-queries-generated.go
```diff
@@ -0,0 +1,69 @@
+// Code generated by github.com/Khan/genqlient, DO NOT EDIT.
```
benjaminjkraft, line 1: I think this may also be an artifact of some testing you did -- can you try removing it and then seeing if snapshot tests still pass?  I could be wrong; if it's needed then no problem!

graphql/client.go
```diff
@@ -40,9 +40,8 @@ type Client interface {
 	// (Errors are returned.) But again, MakeRequest may customize this.
```
benjaminjkraft, line 40: The documentation here will need some updating!  Mostly you can move the descriptions of `query` etc. down into `Payload`, and `retval` down into `Response`, and just replace them with something brief about request and response.

graphql/client.go
```diff
@@ -75,65 +74,64 @@ type Doer interface {
 	Do(*http.Request) (*http.Response, error)
 }
 
-type payload struct {
+// Payload contains all the values required to build queries executed by
+// the graphql.Client.
+type Payload struct {
```
benjaminjkraft, line 79: Now that this is public, we probably ought to name it `Request`?  Since that's what it is.  Not sure why I originally called in `Payload`.

graphql/client.go
```diff
@@ -75,65 +74,64 @@ type Doer interface {
...
 			respBody = []byte(fmt.Sprintf("<unreadable: %v>", err))
 		}
-		return fmt.Errorf("returned error %v: %s", resp.Status, respBody)
+		return fmt.Errorf("returned error %v: %s", httpResp.Status, respBody)
 	}
 
-	var dataAndErrors response
-	dataAndErrors.Data = retval
-	err = json.NewDecoder(resp.Body).Decode(&dataAndErrors)
+	err = json.NewDecoder(httpResp.Body).Decode(resp)
```
benjaminjkraft, line 129: Much better names, thanks!

benjaminjkraft (changes requested):
This looks great, thanks for writing it and for the clear commit message and documentation.

I have a couple of small comments inline; the most important thing is the changes in the API surface `graphql/client.go`, where we just want to be a little picky to get everything right before it's a breaking change.  Everything else is just a suggestion.

Then the last thing is if you could add a changelog entry (under the "vNext" section at the top) -- or actually two; one for the new feature and one for the breaking change to the `MakeRequest` interface (you can crib from the similar one in v0.2.0).

generate/testdata/queries/foo.go
janboll: Oups. going to remove it

generate/testdata/snapshots/TestGenerateWithConfig-ClientGetterNoContext#01-testdata-queries-generated.go
janboll: Indeed. deleted.

graphql/client.go
janboll: That sounds reasonable. Changed it.

generate/operation.go.tmpl
janboll: done

generate/operation.go.tmpl
janboll: Good idea, done!

benjaminjkraft:
Great, thanks for the changes and for the contribution!  I put in a few tiny tweaks to the documentation, which reminded me that this also fixes #19.

benjaminjkraft (approved):

PR #186, opened 2022-04-08
Add support for client that uses GET as transport mechanism

salman-rb:
Current implementation always uses POST as the transport mechanism. Adding GET support enables usage of GET queries for caching simply via URL.

Some notes:
- I left the existing API for creating a new client as is, but the implementation could be much cleaner by introducing some sort of configuration struct when creating a new client
- The construction of the query parameters follows the logic from Apollo's client implementation, which can be found here https://github.com/apollographql/apollo-client/blob/8beb4820edc6352996e08f7f73bde3573f1eb666/src/link/http/rewriteURIForGET.ts
- Updated integration tests to use both sets of clients. Updating the tests to use a test suite would be cleaner

I have:

### DIFF
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -18,6 +18,24 @@ resp, err := getUser(ctx, client, "benjaminjkraft")
 fmt.Println(resp.User.Name, err)
 ```
 
+This request will be sent via an HTTP GET request, with the query, operation name and variables encoded in the URL.
+
+For example, if the query is defined as:
+
+```graphql
+query getUser($login: String!) {
+  user(login: $login) {
+    name
+  }
+}
+```
+
+The URL requested will be:
+
+`https://api.github.com/graphql?operationName%3DgetUser%26query%3D%0Aquery%20getUser(%24login%3A%20String!)%20%7B%0A%20%20user(login%3A%20%24login)%20%7B%0A%20%20%20%20name%0A%20%20%7D%0A%7D%0A%26variables%3D%7B%22login%22%3A%22benjaminjkraft%22%7D`
+
+The client does not support mutations, and will return an error if passed a request that attempts one.
+
 ### … use an API that requires authentication?
 
 When you call `graphql.NewClient`, pass in an HTTP client that adds whatever authentication headers you need (typically by wrapping the client's `Transport`).  For example:
--- a/docs/FAQ.md
+++ b/docs/FAQ.md
@@ -8,6 +8,16 @@ This document describes common questions about genqlient, and provides an index 
 
 There's a [doc for that](INTRODUCTION.md)!
 
+##  … use GET requests instead of POST requests?
+
+You can use `graphql.NewClientUsingGet` to create a client that will use query parameters to create the request. For example:
+```go
+ctx := context.Background()
+client := graphql.NewClientUsingGet("https://api.github.com/graphql", http.DefaultClient)
+resp, err := getUser(ctx, client, "benjaminjkraft")
+fmt.Println(resp.User.Name, err)
+```
+
 ### … use an API that requires authentication?
 
 When you call `graphql.NewClient`, pass in an HTTP client that adds whatever authentication headers you need (typically by wrapping the client's `Transport`).  For example:
--- a/graphql/client.go
+++ b/graphql/client.go
@@ -4,10 +4,12 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"net/url"
+	"strings"
 
 	"github.com/vektah/gqlparser/v2/gqlerror"
 )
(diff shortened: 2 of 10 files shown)


docs/FAQ.md
```diff
@@ -8,6 +8,16 @@ This document describes common questions about genqlient, and provides an index
 
 There's a [doc for that](INTRODUCTION.md)!
 
+##  … use GET requests instead of POST requests?
+
+You can use `graphql.NewClientUsingGet` to create a client that will use query parameters to create the request. For example:
```
csilvers, line 13: It's probably worthwhile to give an example of what this sends over the wire to the server, since it's not obvious what the server would need to do to handle such a request -- unless there's a standard for how to do graphql over GET?  I know there's a standard for POST but I don't know about GET.  (If there is a standard for GET, it may be worth linking to it.)

graphql/client.go
```diff
@@ -53,10 +54,27 @@ type client struct {
...
 // Transport to add those headers.  See example/caller.go for an example.
 func NewClient(endpoint string, httpClient Doer) Client {
+	return newClient(endpoint, httpClient, http.MethodPost)
+}
+
+// NewClientUsingGet returns a Client which makes requests to the given endpoint,
+// suitable for most users.
+//
+// The client makes GET requests to the given GraphQL endpoint using standard
+// GraphQL HTTP-over-JSON transport.  It will use the given http client, or
```
csilvers, line 64: I don't think this sentence is accurate for GET -- it's not using JSON in this case.  This is a great place to describe what query params you pass to the server.

csilvers (commented):
The code looks right to me!, though I'm not one of the gqlgen experts.  Your changes are clear and make a lot of sense.

I think the main issue here is more philosophical, about whether we want to expose this at all.  It's really easy for variables to include personal information, such as passwords, and exposing those via GET -- where they're logged, if nothing else -- is potentially problematic.  There are certainly legit uses for graphql over GET, but I'm worried this introduces an easy-to-use footgun.  But arguably it's not our job as library authors to dictate that kind of thing.

docs/FAQ.md
salman-rb: Its here, will link to it

https://graphql.org/learn/serving-over-http/#get-request

graphql/client.go
salman-rb: Good call out, will update

salman-rb:
> The code looks right to me!, though I'm not one of the genqlient experts. Your changes are clear and make a lot of sense.
> 
> I think the main issue here is more philosophical, about whether we want to expose this at all. It's really easy for variables to include personal information, such as passwords, and exposing those via GET -- where they're logged, if nothing else -- is potentially problematic. There are certainly legit uses for graphql over GET, but I'm worried this introduces an easy-to-use footgun. But arguably it's not our job as library authors to dictate that kind of thing.

Definitely agree with your concerns. I believe the mechanism should be included given that its part of the graphql specification and there are good use cases for it. Maybe there is an opportunity to make it more specific at request time that you are using GET as the transport mechanism to avoid the foot gun, as where the client is first created might be different than where it gets used. I believe this would require changing some of the public APIs to include the request method, will wait for more feedback before making a change like that

benjaminjkraft:
Thanks.  I think this makes sense to include; while users can bring their own client this one is probably common enough it's worth implementing it once, correctly, so everyone can use.

One question is whether it's a good idea to allow GET to be used for mutations.  I would think if you're doing that you're really in for trouble, whether as Craig says via logging your passwords, or simply in something not knowing the request isn't idempotent.  If someone really wants the foot-gun, they can just copy the implementation and remove the check.  Since the client has access to the query you can implement this by just checking if the first word (after whitespace) is "query" or "mutation" (I believe genqlient will never put a comment before the query it sends to the server, although that's perhaps slightly fragile; if it's ever an issue we could give the client the operation type or even the entire parsed query).  We'd want to mention that in the documentation -- actually it's probably a good idea as a recommendation even if we don't require it.

FWIW I actually like having two separate client constructors, as it is now.  It's possible it will become unwieldy, but I feel they're actually separate methods of serializing GraphQL over HTTP, not just two different request methods.

Lastly, I think you still need to sign the CLA?  Or at least, you didn't check the checkbox :-) .

docs/FAQ.md
benjaminjkraft: Yeah, it's semi-standard.  (The official spec has almost nothing about wire format, but the `graphql-js` and Apollo implementations are effectively the standard.)  May as well include the link but in practice I think most servers will either support it or not.

salman-rb: Didn't link directly (not sure if this will be stable) but described the format in the FAQ

benjaminjkraft:
Great! Looks good to merge as soon as the CLA is sorted, let us know if you need anything from us on that!

salman-rb:
> Great! Looks good to merge as soon as the CLA is sorted, let us know if you need anything from us on that!

Should be sorted! Let me know if there are any issues

PR #192, opened 2022-05-10
Allow absolute paths in schema files.

csilvers:
Thsi is useful for some out-of-tree testing I want to do.  It's
unfortunate (imo) that filepath.join doesn't have this behavior by
default.

Test plan:
go test ./...

I have:

### DIFF
--- a/generate/config.go
+++ b/generate/config.go
@@ -51,6 +51,16 @@ type TypeBinding struct {
 	Unmarshaler       string `yaml:"unmarshaler"`
 }
 
+// pathJoin is like filepath.Join but 1) it only takes two argsuments,
+// and b) if the second argument is an absolute path the first argument
+// is ignored (similar to how python's os.path.join() works).
+func pathJoin(a, b string) string {
+	if filepath.IsAbs(b) {
+		return b
+	}
+	return filepath.Join(a, b)
+}
+
 // ValidateAndFillDefaults ensures that the configuration is valid, and fills
 // in any options that were unspecified.
 //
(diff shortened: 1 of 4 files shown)


generate/config.go
```diff
@@ -51,6 +51,16 @@ type TypeBinding struct {
 	Unmarshaler       string `yaml:"unmarshaler"`
 }
 
+// pathJoin is like filepath.Join but 1) it only takes two argsuments,
+// and b) if the second argument is an absolute path the first argument
+// is ignored (similar to how python's os.path.join() works).
+func pathJoin(a, b string) string {
```
StevenACoffman, line 57: I would suggest a more accurate name such as `func abs(basedir, path string) string {`
This is then similar to the stdlib's `filepath.Abs` but allows you to specify the base directory instead of using the current working directory.

func [Abs](https://cs.opensource.google/go/go/+/go1.18.2:src/path/filepath/path.go;l=243) [¶](https://pkg.go.dev/path/filepath#Abs)
func Abs(path [string](https://pkg.go.dev/builtin#string)) ([string](https://pkg.go.dev/builtin#string), [error](https://pkg.go.dev/builtin#error))
Abs returns an absolute representation of path. If the path is not absolute it will be joined with the current working directory to turn it into an absolute path. The absolute path name for a given file is not guaranteed to be unique. Abs calls Clean on the result.

csilvers: `abs` isn't very accurate because there's no requirement that `a` be an absolute path.  I really see this as a "proper" path-join (and the default go implementation as not-proper).  But ymmv.

StevenACoffman: This is an unexported helper, so you do you.

generate/config.go
```diff
@@ -59,14 +69,14 @@ type TypeBinding struct {
...
 	c.baseDir = baseDir
 	for i := range c.Schema {
-		c.Schema[i] = filepath.Join(baseDir, c.Schema[i])
+		c.Schema[i] = pathJoin(baseDir, c.Schema[i])
 	}
 	for i := range c.Operations {
-		c.Operations[i] = filepath.Join(baseDir, c.Operations[i])
+		c.Operations[i] = pathJoin(baseDir, c.Operations[i])
 	}
 	c.Generated = filepath.Join(baseDir, c.Generated)
```
benjaminjkraft, line 77: May as well do Generated too!

generate/config.go
benjaminjkraft: I agree the name isn't ideal, but I don't have a better idea, so I'm fine with this. (Closest thing I can think of in the stdlib is `net/url.ResolveReference` which I don't think is the greatest name either.)

benjaminjkraft (approved):
Cool, thanks!

PR #195, opened 2022-05-12
Reject operation or argument names that are Go keywords

### LINKED ISSUE #194: Generate fails when building a mutation that takes enum as input.
**Describe the bug**
Generate fails when building a mutation that takes enum as input.

**To Reproduce**
The following fails to generate: 
```graphql
# Schema snippet
enum IPAddressType {
  v4
  v6
}
```
```graphql
# Query snippet
mutation AllocateIpAddress($app: ID!, $region: String, $type: IPAddressType) {
    allocateIpAddress(input: {
        appId: $app,
        region: $region,
        type: $type
    }) {
        app {
            name
        }
    }
}
```
```shell
genqlient internal error: failed to gofmt code:
	680:5: expected 'IDENT', found 'type' (and 8 more errors)
	680:5: expected 'IDENT', found 'type' (and 8 more errors)
	680:5: expected 'IDENT', found 'type' (and 8 more errors)
	680:5: expected 'IDENT', found 'type' (and 8 more errors)
	680:5: expected 'IDENT', found 'type' (and 8 more errors)
	680:5: expected 'IDENT', found 'type' (and 8 more errors)
	680:5: expected 'IDENT', found 'type' (and 8 more errors)
	680:5: expected 'IDENT', found 'type' (and 8 more errors)
	680:5: expected 'IDENT', found 'type' (and 8 more errors)---source code---
```
The following slight change generates without issue: 
```graphql
mutation AllocateIpAddress($app: ID!, $region: String) {
    allocateIpAddress(input: {
        appId: $app,
        region: $region,
        type: v4
    }) {
        app {
            name
        }
    }
}
```

**Expected behavior**
The generator should successfully complete. 

**genqlient version**
0.4.0

benjaminjkraft:
GraphQL is pretty restrictive about its identifiers, so for the most
part we can and do safely use GraphQL identifiers in the Go we generate
with attention only to conflicts with other such identifiers. But we do
need to check one thing, which is that the identifier isn't a Go
keyword. (If it is, the generated code will almost certainly fail to
compile, but often with a confusing error message.) In this commit I add
such checks.

The most likely place to run into trouble here is argument names, which
are often one word and are used as-is.  Operation names, if unexported,
can also be keywords, although in practice they're usually multiword.
Field names are always exported, thus safe.  Generated type names are
always prefixed, camel-cased, with the operation name, so they always
contain an uppercase letter (even if the operation name is lowercase),
but type-names specified by `typename` may collide, so we check those.
In theory we could check type-names specified by `bind`, but these must
be defined by the user, so their code will already fail to compile, so I
didn't bother.  I think that's all the places to consider, although it's
hard to be sure.  In summary, we check argument names, operation names,
and user-specified type names.

Fixes #194.

Test plan: make check

I have:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -33,6 +33,8 @@ When releasing a new version:
 
 ### Bug fixes:
 
+- genqlient now explicitly rejects argument, operation, and type names which are Go keywords, rather than failing with an opaque error.
+
 ## v0.4.0
 
 Version 0.4.0 adds several new configuration options, as well as additional methods to simplify the use of interfaces.
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -164,6 +164,10 @@ func (g *generator) convertArguments(
 	name := "__" + operation.Name + "Input"
 	fields := make([]*goStructField, len(operation.VariableDefinitions))
 	for i, arg := range operation.VariableDefinitions {
+		if goKeywords[arg.Variable] {
+			return nil, errorf(arg.Position, "variable name must not be a go keyword")
+		}
+
 		_, options, err := g.parsePrecedingComment(arg, nil, arg.Position, queryOptions)
 		if err != nil {
 			return nil, err
@@ -317,6 +321,9 @@ func (g *generator) convertDefinition(
 	// Determine the name to use for this type.
 	var name string
 	if options.TypeName != "" {
+		if goKeywords[options.TypeName] {
+			return nil, errorf(pos, "typename option must not be a go keyword")
+		}
 		// If the user specified a name, use it!
 		name = options.TypeName
 		if namePrefix != nil && namePrefix.head == name && namePrefix.tail == nil {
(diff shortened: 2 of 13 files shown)


csilvers (approved):
Good thinking!

So sad I won't be able to use `goto` as a graphql variable name...

benjaminjkraft:
Never fear, you can still use `goTo`!

JuanitoFatas:
Sorry to ask here. I was hitting by this and wonder what if a mutation requires an input that is the same name as Go Keywords (e.g. `type`), is there a way to get around this?

```graphql
mutation Mutation {
  resourceCreate(name: "Test", type: "type-1") {
    resource {
      name
      type
    }
  }
}
```

benjaminjkraft:
That query as written should work, I believe. Of course the query you actually want is probably something like
```graphql
mutation Mutation($name: String, $type: String) {
  resourceCreate(name: $name, type: $type) {
    resource {
      name
      type
    }
  }
}
```
In such a case there's no restriction on the variable name -- you could just as easily name it `$typ` or `$type_`. If you are still having problems, please open a new issue with a reproduction of the problem!

PR #196, opened 2022-05-15
Simplify errors tests a bit so they don't all have to write a schema

benjaminjkraft:
Some of the errors tests need to have their own schema, so the schema
can do something weird (or even be entirely invalid!).  But most can
still share a schema.  In this commit I have those indeed share a
schema, to avoid having to have a bunch of copies of mostly the same
schema.  While doing so I noticed one error whose location wasn't very
useful, and fixed it.

Test plan: make check

I have:


csilvers (approved):
Nice!

PR #197, opened 2022-05-15
Validate against a case gqlparser doesn't catch

### LINKED ISSUE #176: genqlient crash in preprocessQueryDocument
Stacktrace:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x40 pc=0x6d064d]

goroutine 1 [running]:
github.com/Khan/genqlient/generate.(*generator).preprocessQueryDocument.func1(0xc000174240, 0xc0000dac00)
	/home/benkraft/.go/pkg/mod/github.com/!khan/genqlient@v0.4.0/generate/generate.go:188 +0x2d
[...]
```
which is here:
```go
		fieldType := g.schema.Types[field.Definition.Type.Name()]
```
I assume some particular case gets us here before validation, or validation doesn't catch something it should.  Anyway, we should give a proper error that the type is undefined.

benjaminjkraft:
While writing tests at some point I came across an invalid query that
gqlparser doesn't catch (see vektah/gqlparser#221), and which causes
a panic for us.  Now we validate for it and return a nice error instead of
panicking.

Fixes #176.

Test plan: make check

I have:

### DIFF
--- a/generate/generate.go
+++ b/generate/generate.go
@@ -224,16 +224,40 @@ func (g *generator) preprocessQueryDocument(doc *ast.QueryDocument) {
 	validator.Walk(g.schema, doc, &observers)
 }
 
+// validateOperation checks for a few classes of operations that gqlparser
+// considers valid but we don't allow, and returns an error if this operation
+// is invalid for genqlient's purposes.
+func (g *generator) validateOperation(op *ast.OperationDefinition) error {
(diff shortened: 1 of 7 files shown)


StevenACoffman (approved):

csilvers (approved):
Heh!

Issue #178, opened 2022-03-17
Option to use pointers for optional fields


connec:
**Is your feature request related to a problem? Please describe.**

I'm working on a Go service that needs to speak to a number of 'backend' GraphQL APIs. I'm trying to get to a setup where both sides of the communication use generated types, in order to bring integration issues to compile-time.

I'm using `gqlgen` for the server, which defaults to using pointers to describe optional fields. For example, given a GraphQL type such as:

```graphql
type User {
    id: ID!
    name: String!
    avatarUrl: String
}
```

`gqlgen` will generate a server-side model like:

```go
type User struct {
	AvatarURL *string `json:"avatarUrl"`
	ID string `json:"id"`
	Name string `json:"name"`
}
```

E.g., the `AvatarURL` field is a `*string` since it is optional in the schema.

However, when I use `genqlient` I get a model like:

```go
type userGetByIDUser struct {
  AvatarUrl string `json:"avatarUrl"`
  Id string `json:"id"`
  Name string `json:"name"`
}
```

I'm sure reasonable people could disagree about how optional fields should be encoded in Go, but I'm personally a fan of pointers for this purpose, at least as a default (I'd rather assume `""` can be a valid, present `string` than treat it as `<none>` by default). In this case, it's particularly jarring since the server-side models (worked on by the same engineers) use a different paradigm.

From the looks of things, annotating optional fields explicitly with `pointer: true` would generate the models I desired, but for me this undermines the advantages of using code generation to ensure consistency between the client and server – it's something that can be forgotten.

**Describe the solution you'd like**

I'd like a configuration for `genqlient.yaml` that will lead to using pointers for optional fields in generated types. E.g. I'd like `genqlient` to generate the following, without additional annotations:

```go
type userGetByIDUser struct {
  AvatarUrl *string `json:"avatarUrl"`
  Id string `json:"id"`
  Name string `json:"name"`
}
```

After starting to write this I found the note on [Optionality and pointers](https://github.com/Khan/genqlient/blob/main/docs/DESIGN.md#optionality-and-pointers) in `DESIGN.md`, and while it's a fine take, it would be really useful to be able to configure this 'globally' when working with different design decisions.

**Describe alternatives you've considered**

Annotating all optional fields is an obvious "functional" solution, but it undermines some of the goals I have for using code generation.

I'm also considering forking, to change the default for our use, but it would be far preferable to have upstream support. I may be able to work on a PR, if one would be considered.

[...]

PR #198, opened 2022-05-18
wip: Add support for mapping all nullable types as pointers

connec:
This is a bare minimum implementation of the approach suggested in https://github.com/Khan/genqlient/issues/178#issuecomment-1077559343, which I figured I'd open sooner than later for early feedback!

Per the new detail in the sample `docs/genqlient.yaml`:

```yaml
# Customize how optional fields are handled.
optional:
  # Customize how models are generated for optional fields. This can currently
  # be set to one of the following values:
  # - value (default): optional fields are generated as values, the same as
  #   non-optional fields. E.g. fields with GraphQL types `String` or `String!`
  #   will both map to the Go type `string`. When values are absent in
  #   responses the zero value will be used.
  # - pointer: optional fields are generated as pointers. E.g. fields with
  #   GraphQL type `String` will map to the Go type `*string`. When values are
  #   absent in responses `nil` will be used.
  output: value
```

The documentation is slightly misleading atm as there's actually no validation of the values, and only `pointer` will have any effect (anything else has the default behaviour). It also affects arguments, no just fields, which is desirable imo. There are also no tests, as I'm not quite sure where best to put them 😅 Any pointers (pun intended?) would be great.

I've tested this out with a small local setup, which isn't fully integrated yet, so I'm not at all confident that it fully works. It does generate what I'd expect for a simple schema, e.g. given:

```graphql
# schema.graphql
scalar DateTime

type Query {
  user(id: ID!): User
}

type User {
  avatarUrl: String
  brandID: ID
  brandVisible: Boolean
  createdAt: DateTime
  firstName: String
  id: ID!
  lastName: String
  name: String
  primaryEmail: String
}
```

And:

```graphql
# operations.graphql
query userGetByID($id: ID!) {
  user(id: $id) {
    avatarUrl
    brandID
    brandVisible
    createdAt
    firstName
    id
    lastName
    name
    primaryEmail
  }
}
```

It will generate:

```go
func userGetByID(
	ctx context.Context,
	client graphql.Client,
	id string,
) (*userGetByIDResponse, error) {
	/* ... */
}

type userGetByIDResponse struct {
	User *userGetByIDUser `json:"user"`
}

type userGetByIDUser struct {
	AvatarUrl *string `json:"avatarUrl"`
	BrandID *string `json:"brandID"`
	BrandVisible *bool `json:"brandVisible"`
	CreatedAt *time.Time `json:"createdAt"`
	FirstName *string `json:"firstName"`
	Id string `json:"id"`
	LastName *string `json:"lastName"`
	Name *string `json:"name"`
	PrimaryEmail *string `json:"primaryEmail"`
}

// ...
```

---

I have:


StevenACoffman:
Hi! The majority of gqlgen users are using versions where resolvers return pointers, so I think that should be the default now and people can opt-in to the value behavior.

Otherwise, it's kind of a jarring change.

benjaminjkraft:
This seems like a great start, thank you!  Fingers crossed that the tests don't find any place it needs to be more complicated.

> It also affects arguments, no just fields, which is desirable imo.

If that's simpler, and seems to work for you, I think it's fine!  Only thing is to rename the option to match.  Maybe just `optional: pointer` and later if we add more knobs that can be a shorthand for something else.

> There are also no tests, as I'm not quite sure where best to put them

Hmm, slightly awkward actually.  In general this sort of thing goes in `TestGenerateWithConfig`, but you'll definitely need to test it on a more complex query than it currently uses to get any interesting behavior.  (Currently it uses `generate/testdata/queries/SimpleQuery.graphql`, the schema it uses is in `generate/testdata/queries/schema.graphql`.  This writes out to various snapshots in `generate/testdata/snapshots`, as well as checking the snapshots compile.)  Probably the easiest thing is just to add the query to use as a parameter to the table-driven test (currently it's just a constant `sourceFilename`), so you can add a few tests using more complicated queries (or use some of the existing ones in the same directory, which are probably plenty complex).

Then the other thing is whether we need an integration test.  If we do, it will be a bit annoying, since this is a global config option and those all share one config.  But if we are able to stick to doing exactly the same thing as the `pointer: true` option, that may not be necessary.

connec:
Thanks for the fast feedback! I'll reconsider the naming and investigate and adding a test, though it will likely be a couple of days before I get to it.

connec:
I've added a couple of tests, one that has a few nullable outputs (including lists) and one with some nullable inputs. The result is what I expect.

benjaminjkraft:
Neat, thanks for putting this together!

Only thing I notice is that a value of type `[String]` becomes `[]*string`, not `*[]*string`, which in theory might be what one would expect.  But I think that's probably fine, since slices already have distinct nil vs. empty slices, and it's more consistent with how the `pointer` option already works.  Which is for the best because it might be quite annoying to implement (would require redoing the handling of such types in marshalers, which is necessarily very tricky code to work with).  I'm gonna add a few words to the documentation to mention that, and then merge.

benjaminjkraft (approved):

connec:
Awesome! I did think about `[String]` and `*[]*string`. I could have gone either way but as you say `[]T` can already distinguish between `nil` and empty. I think this matches what gqlgen does by default as well.

PR #209, opened 2022-07-27
make output deterministic for graphql interfaces

vikstrous2:
We noticed that the output from genqlient can be non-deterministic when a graphql query queries a field that's an interface. This PR fixes that by sorting the types as soon as we extract them from the schema.

The snapshots for the tests need to be updated, but I don't know an easy way to do that. Is there a command or a flag that I can just flip somewhere to update them?

I have:
- n/a Included a link to the issue fixed, if applicable
- n/a Included documentation, for new features

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -26,6 +26,8 @@ When releasing a new version:
 
 ### Bug fixes:
 
+- Fixed non-deterministic generated code when querying graphql interfaces
+
 ## v0.5.0
 
 Version 0.5.0 adds several new configuration options and convenience features. Note that genqlient now requires Go 1.16 or higher, and is tested through Go 1.18.
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -11,6 +11,7 @@ package generate
 
 import (
 	"fmt"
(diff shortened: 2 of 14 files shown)


StevenACoffman:
@vikstrous2 Looks like some test failures

benjaminjkraft:
Huh, I'm a bit surprised since we've never run into this in the tests (or at Khan as far as I know). I'm guessing the issue is actually if you have multiple schema files and the order they get read in is nondeterministic, or something. But anyway, it makes good sense to make this more stable even if in principle we could solve it elsewhere, so I'm happy to merge once tests are updated.

For the tests, it should just be `UPDATE_SNAPSHOTS=1 make check`.

StevenACoffman:
Yeah, I ran  `UPDATE_SNAPSHOTS=1 make check` and then pushed the changed version to this PR to clarify which test failures were not fixed by that.

benjaminjkraft:
Ah! It looks like we also just need `go generate ./...` to update the genqlient-generated files for integration tests.

benjaminjkraft:
Ok, there we go!

PR #219, opened 2022-08-15
Bump golangci-lint and (max) Go versions

### LINKED ISSUE #216: Add 1.19 to tests
Go 1.19 is out, so we may as well test against it!

Note this also means we can look at dropping 1.16 and 1.17 (since Go officially only supports the latest two versions), although as usual my preference is to keep them until there's a reason to remove them. From a quick look I don't see anything exciting about dropping 1.16; dropping 1.17 of course gets us generics which may have some value internally but with which there's probably some value in being a bit conservative at first. (1.18 also adds fuzzing, but when I tried to use it I found it ended up mostly just fuzzing the parser; and anyway we could just run that only on 1.18+.)

benjaminjkraft:
Lint is failing with some inscrutable panic (on a commit with no code
changes). Let's try bumping the version in case they fixed it.

Additionally, the new version's github action uses Go 1.19, which means
it pulls in gofmt updates to match the new [doc-comment formatting rules][1].
So I added Go 1.19 to our list of versions to test (fixes #216)
and updated some of our doc-comments to format better in the
new world (mostly using the new link syntax).

[1]: https://go.dev/doc/comment

Test plan: make lint

I have:


benjaminjkraft:
Turns out the newest golangci-lint also uses Go 1.19, and turns out that means we need to update gofmt which does some new doc-comment formatting, and then I want to tweak some of our formatting while I'm at it. Will return to this later.

csilvers (approved):
Looks good once the latest lint errors are fixed!

Issue #220, opened 2022-08-15
running generate with package_bindings causes an error (@main)


penina-nyt:
**Describe the bug**
received the following error trying to generate code
```
genqlient internal error: failed to gofmt code:
	16:21: expected ';', found '-' (and 5 more errors)
	16:21: expected ';', found '-' (and 5 more errors)
	16:21: expected ';', found '-' (and 5 more errors)
	16:21: expected ';', found '-' (and 5 more errors)
	16:21: expected ';', found '-' (and 5 more errors)
	16:21: expected ';', found '-' (and 5 more errors)---source code---
```

**To Reproduce**
in `genqlient.yaml`, specify the path to your models.
```
package_bindings:
- package: path/to/models.go
```
run `go run github.com/Khan/genqlient@main` with a defined query file. 

**Expected behavior**
Should not fail and should successfully generate the file reusing the project's existing models 

**genqlient version**
github.com/Khan/genqlient@main 

**Additional context**
Here's a snipped from what was printed on the command line:
```
   1 | // Code generated by github.com/Khan/genqlient, DO NOT EDIT.
   2 | 
   3 | package client
   4 | 
   5 | import (
   6 | 	"command-line-arguments"
   7 | 	"context"
   8 | 	"github.com/Khan/genqlient/graphql"
   9 | )
  10 | 
  11 | 
  12 | 
  13 | 
  14 | // Type is returned by Query on success.
  15 | type Type struct {
  16 | 	Field command-line-arguments.Field `json:"field"`
> 17 | }
```

benjaminjkraft:
Weird! Something seems to think that your models are in a package named `command-line-arguments`. Does that string by chance appear anywhere in your genqlient config or models file? (It doesn't seem to appear in the genqlient source.) If not, are you able to post a bit more of your config or ideally a sample config and models file that reproduces the issue?

benjaminjkraft:
Whoa, weird, nice find!

It took me a moment to notice but I think the problem here is that the value for `package:` is supposed to be a package-path, not a local path to a file. (Then the `go/packages` API is I guess assuming you want to compile that file as a standalone package and that somehow works but results in a bogus package-path.)

That is, it should be:
```yaml
package_bindings:
- package: github.com/yourorg/yourrepo/path/to
```
Does it work if you change it like that? (And if that's the issue, is there somewhere in the genqlient documentation you got that from, or that we could be clearer where you would have seen it? We should definitely look for this and give a better error, too!)

[...]

PR #221, opened 2022-08-15
Give a better error if package_bindings is a file

benjaminjkraft:
The entries under the new `package_bindings` field should be packages,
but it's an easy mistake to put a file path instead (most of the other
fields in `genqlient.yaml` are files). Due to some bizzare behavior from
`go/packages` (described in #220), if you do that you get weird broken
code that gives you no clue what is wrong. Instead, let's guess if what
you gave us looks like a filename, and report a nice error if so.

Test plan: crossed fingers

I have:


csilvers (approved):
Makes sense to me.  Don't go naming your directories 'something.go/' though!

benjaminjkraft:
I think in theory it's legal! As long as your package clause is still `package something`. But it seems unlikely anyone actually does that.

PR #222, opened 2022-09-01
Fix `optional` default value

knao124:
## changes
- fix `optional` default value in docs/genqlient.yaml

## issue
current yaml causes an unmarshal error like below

```
line 14: cannot unmarshal !!map into string
```

see: https://github.com/Khan/genqlient/blob/main/generate/config.go#L34

### DIFF
--- a/docs/genqlient.yaml
+++ b/docs/genqlient.yaml
@@ -98,20 +98,19 @@ use_struct_references: boolean
 use_extensions: boolean
 
 # Customize how optional fields are handled.
-optional:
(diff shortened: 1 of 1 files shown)


benjaminjkraft:
Thanks for catching this! It looks good to me; could you please sign the [Khan Academy CLA](https://www.khanacademy.org/r/cla)?

csilvers:
@knao124 let us know when you've signed the CLA!

benjaminjkraft:
This is easy enough to fix I'll just redo it to avoid CLA issues.

PR #238, opened 2022-11-18
Expose the graphql operation as a constant in the generated genqlient file

### LINKED ISSUE #236: Export the query-over-the-wire as a constant
Generated genqlient code might look like this:
```
func Progress_AssignmentByID(
        ctx gqlclient.KAContext,
        client graphql.Client,
        id string,
) (*Progress_AssignmentByIDResponse, error) {
        req := &graphql.Request{
                OpName: "Progress_AssignmentByID",
                Query: `
query Progress_AssignmentByID ($id: String!) {
        user {
                assignment(id: $id) {
                        studentKaids
                        exerciseConfig {
                                itemPickerStrategy
                                assessmentItemIds
                        }
                }
        }
}
`,
                Variables: &__Progress_AssignmentByIDInput{
                        Id: id,
                },
        }
        var err error
        ...
```

`Request.Query` differs from the graphql code as written in the source because, among other things, the `@genqlient` directive has been removed.

For our application, we'd like to know in Go code what the actual text-over-the-wire is.  We can look in generated/safelist.json, but it would be easier just to expose it in Go.  I'm thinking of an API like this:
```
const Progress_AssignmentByIDQuery = `
query Progress_AssignmentByID ($id: String!) {
        user {
                assignment(id: $id) {
                        studentKaids
                        exerciseConfig {
                                itemPickerStrategy
                                assessmentItemIds
                        }
                }
        }
}
`

func Progress_AssignmentByID(
        ctx gqlclient.KAContext,
        client graphql.Client,
        id string,
) (*Progress_AssignmentByIDResponse, error) {
        req := &graphql.Request{
                OpName: "Progress_AssignmentByID",
                Query: Progress_AssignmentByIDQuery,
                Variables: &__Progress_AssignmentByIDInput{
                        Id: id,
                },
        }
        var err error
        ...
```

Is this a change you would be amenable to seeing?

csilvers:
This allows client code to see the operation (query or mutation) exactly as genqlient sends it over the wire.  This data was already available in the generated safelist.json file, but now it's easily available from Go code as well.
    
Fixes #236

I have:

### DIFF
--- a/generate/operation.go.tmpl
+++ b/generate/operation.go.tmpl
@@ -1,3 +1,5 @@
+const {{.Name}}Operation = `{{$.Body}}`
+
 {{.Doc}}
 func {{.Name}}(
     {{if ne .Config.ContextType "-" -}}
@@ -15,7 +17,7 @@ func {{.Name}}(
 ) (*{{.ResponseName}}, {{if .Config.Extensions -}}map[string]interface{},{{end}} error) {
     req := &graphql.Request{
         OpName: "{{.Name}}",
-        Query:  `{{.Body}}`,
+        Query:  {{.Name}}Operation,
     {{if .Input -}}
         Variables: &{{.Input.GoName}}{
         {{range .Input.Fields -}}
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -25,6 +25,7 @@ When releasing a new version:
 ### New features:
 
 - You can now bind all types from a package in `genqlient.yaml` using the new `package_bindings` option.
+- The graphql operation (query or mutation) as sent over the wire is now exposed via a constant in the generated genqlient .go file.
 
 ### Bug fixes:
 
--- a/example/generated.go
+++ b/example/generated.go
@@ -69,6 +69,15 @@ func (v *getViewerViewerUser) GetMyName() string { return v.MyName }
 // GetCreatedAt returns getViewerViewerUser.CreatedAt, and is useful for accessing the field via an interface.
 func (v *getViewerViewerUser) GetCreatedAt() time.Time { return v.CreatedAt }
 
+const getUserOperation = `
+query getUser ($Login: String!) {
+	user(login: $Login) {
+		theirName: name
+		createdAt
+	}
+}
+`
+
 // getUser gets the given user's name from their username.
 func getUser(
 	ctx context.Context,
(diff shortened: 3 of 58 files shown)


dnerdy (approved):
This looks good to me!

Note that the constant name could conflict with a separate operation with the same name, e.g. getViewer and getViewerOperation (which would have the operation string getViewerOperationOperation). Such a case seems very unlikely, though (and is definitely confusing), so I don't think we need to worry about it.

csilvers:
> This looks good to me!
> 
> Note that the constant name could conflict with a separate operation with the same name, e.g. getViewer and getViewerOperation (which would have the operation string getViewerOperationOperation). Such a case seems very unlikely, though (and is definitely confusing), so I don't think we need to worry about it.

Yeah, the same is currently true of getViewerResponse as well.  I agree we probably don't need to worry about it.

generate/operation.go.tmpl
```diff
@@ -1,3 +1,5 @@
+const {{.Name}}Operation = `{{$.Body}}`
```
benjaminjkraft, line 1: May as well put a doc-comment on this? Probably just saying it's the query executed by `{{.Name}}`.

benjaminjkraft (approved):
Seems reasonable to me!

vikstrous2:
This is causing a bug in our generated code. The name of an interface is the same as the name of this constant. It seems like "Operation" is used as a suffix for some existing identifiers in some cases. Maybe you can use the suffix "OperationQuery" instead?

vikstrous2:
Oh, maybe this is exactly the case you guys were talking about where we have a query like:
```
query GetX($operationID: String!) {
  operation(operationID: $operationID) {
    ... SomeOperationType
  }
}
```

csilvers:
https://github.com/Khan/genqlient/pull/241

benjaminjkraft:
@vikstrous2 thanks for the report, it seems this case is more likely than we thought. I guess it makes sense when you point out that it can collide with a toplevel field (not just a similarly-named query), where `operation` is a totally reasonable name (probably way more common than `response`).

I merged the change for `{{.Name}}_Operation` to start with. (Thanks @csilvers for the quick fix!) I think that's fine if we don't come up with something better. Another option is to give it a prefix, e.g. `RawOperationFor{{.Name}}`, in hopes that that's a less likely collision, although then it doesn't sort as nicely. A weirder idea would be to put the query in a map or struct or something; then we just have to reserve one name globally. But that may have negative impacts if the operation-body can no longer be a const.

Ideally we'd do the same for Response, but that's probably not worth it at this point. If it comes up with Response we can make the pattern/suffix configurable. (Actually, another solution would be to put this behind an option which would also configure the pattern.)

PR #243, opened 2022-12-03
Automatically update generated files if UPDATE_SNAPSHOTS=1

### LINKED ISSUE #212: Auto-update generated files if `UPDATE_SNAPSHOTS=1` is set
If you do `UPDATE_SNAPSHOTS=1 go test ./...` that will:
1. run the snapshot tests, updating any changed snapshots
2. run the integration tests and (if you have tokens) example tests
3. check that the code for the integration tests and example tests is up-to-date

Step 3 is not strictly a snapshot test; the generated code is actually checked in and used. But, I mean, it's basically the same! So we should probably update it in the same places; if you need to update the snapshots you also probably need to update the integration tests. It's a little weirder because you'll also now need to re-run them (since they test things other than the snapshots) but that's probably ok.

benjaminjkraft:
If you do `UPDATE_SNAPSHOTS=1 go test ./...` that will:
1. run the snapshot tests, updating any changed snapshots
2. run the integration tests and (if you have tokens) example tests
3. check that the code for the integration tests and example tests is up-to-date

Step 3 is not strictly a snapshot test; the generated code is actually checked in and used. But, I mean, it's basically the same! So now, we also update it if you asked to update snapshots, which is hopefully a little more convenient.

Fixes #212.

Test plan:
Make a trivial change to `example/generated.go`; `go test ./...` should now fail. `UPDATE_SNAPSHOTS=1 go test ./...` should also fail, but say it updated the snapshot, and the change should be reverted. Run `go test ./...` again; it should pass again.

I have:

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -40,9 +40,7 @@ Notes for contributors:
 - If `GITHUB_TOKEN` is available in the environment, it also checks that the example returns the expected output when run against the real API.  This is configured automatically in GitHub Actions, but you can also use a [personal access token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) with no scopes.  There's no need for this to cover anything in particular; it's just to make sure the example in fact works.
 - Tests should use `testify/assert` and `testify/require` where convenient (when making many simple assertions).
 
-If you update any code-generation logic or templates, even if no new tests are needed you'll likely need to:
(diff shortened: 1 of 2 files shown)


csilvers (approved):
Makes sense to me!

StevenACoffman (approved):
Yay! @kevinbarabash This is the behavior I think you were looking for in webapp, right?

Issue #251, opened 2023-02-09
New generic type option for nullable values


DylanRJohnston-FZ:
**Is your feature request related to a problem? Please describe.**
Currently working on a codebase where we don't like using pointers for null values, but small types like booleans present problems with using the `zero` value for when it's absent. False is a perfectly valid value for a partial update of a boolean field in a larger struct.

**Describe the solution you'd like**
We have a custom built a generic `optional` value much you'd see in languages like Rust, Java, or Haskell that we'd like to use instead. I propose adding a third option to the `optional` configuration, something like `generic`, and then another configuration option much like `context_type`, `optional_generic_type` that can be used in place of pointers.

```
optional: generic
optional_generic_type: optional.Value
```

The question is whether to have the generic value without its generic parameter, or include some placeholder like `optional.Value[A]` or `optional.Value[%]`.

Having this as a third option means you won't break compatibility with <1.18 unless someone opts into it.

**Describe alternatives you've considered**
The available alternatives of zero values, or pointers each have significant drawbacks.

**Additional context**
This is the full implementation of our generic optional type `optional.Value`, but the configuration options allow anyone else to sub their own in. Happen to open a Pull Request with the changes if you like the idea.

```go
package optional

import (
	"encoding/json"
)

type Value[V any] struct {
	value V
	ok    bool
}

func Some[V any](value V) Value[V] {
	return Value[V]{value: value, ok: true}
}

func None[V any]() Value[V] {
	return Value[V]{ok: false}
}

func (v Value[V]) Unpack() (V, bool) {
	return v.value, v.ok
}

func (v Value[V]) Get(fallback V) V {
	if v.ok {
		return v.value
	}

	return fallback
}

func FromPtr[V any](ptr *V) Value[V] {
	if ptr == nil {
		return None[V]()
	}
	
	return Some(*ptr)
}

func (value Value[V]) MarshalJSON() ([]byte, error) {
	if value.ok {
		return json.Marshal(value.value)
	} else {
		return json.Marshal((*V)(nil))
	}
}

func (value *Value[V]) UnmarshalJSON(data []byte) error {
	v := (*V)(nil)

	err := json.Unmarshal(data, &v)
	if err != nil {
		return err
	}

	if v != nil {
		value.value = *v
		value.ok = true
	}

	return nil
}
```

benjaminjkraft:
#252

PR #252, opened 2023-02-10
Add "generic" option to the "optional" configuration for handling nullable types

DylanRJohnston:
This is an implementation for #251, it adds a new `"generic"` option for the `"optional"` configuration, and a companion type `"optional_generic_type"` which is a fully qualified type with a placeholder `%` for the generic parameter.

I have tested this locally with a custom `optional.Value[A]` implementation but I'm unsure how to add tests for this without bumping the Go version from 1.16 to 1.18. I would appreciate any input on how best to achieve this. Interested in any feedback on the design before I fill out the details on documentation and changelog.

I have:


generate/convert.go
```diff
@@ -262,6 +263,13 @@ func (g *generator) convertType(
 		// options work, recursing here isn't as connvenient.)
 		// Note this does []*T or [][]*T, not e.g. *[][]T.  See #16.
 		goTyp = &goPointerType{goTyp}
+	} else if !typ.NonNull && g.Config.Optional == "generic" {
+		qualifiedTypeName := regexp.MustCompile("[^/]+$").FindString(g.Config.OptionalGenericType)
+
```
DylanRJohnston, line 268: I saw that you already have functionality for dealing with fully qualified to just qualified import names inside the `(*generator).ref` method, but it's mixed in with logic around registering imports and so didn't feel appropriate to use here.

generate/generate.go
```diff
@@ -388,6 +388,13 @@ func Generate(config *Config) (map[string][]byte, error) {
 		}
 	}
 
+	if g.Config.Optional == "generic" {
+		_, err = g.ref(g.Config.OptionalGenericType)
+		if err != nil {
+			return nil, err
+		}
+	}
```
DylanRJohnston, line 396: `OptionalGenericType` must be defined if `optional: "generic"` is set, is this the appropriate place for a validation error to be thrown if it's unset?

DylanRJohnston:
Here's an example of it in use https://github.com/DylanRJohnston/genqlient/compare/main...DylanRJohnston:genqlient:example

benjaminjkraft:
I love the concept, thanks very much for implementing it! I don't have time for a proper review today but one thing I see is we'll want tests for this -- there are snapshot tests already set up and you'll just have to make some way to skip these on versions below 1.18. I'm a bit busy right now but I'll try to get back with a full review within a week or so.

benjaminjkraft (commented):
Cool, this all seems reasonable to me once it has tests and a bit of documentation! I don't love the config syntax having two separate options but I don't have any better ideas right now so let's go with it. The only other thing I wonder is whether trying to substitute in a `%` in an arbitrary expression is going to be more confusing than helpful -- for example if you do `path/to.MyWeirdOptional[path/to.MyArg, %]` it probably(?) won't work because we don't know how to do the import. So I wonder if we want to just say it has to be a generic with one argument, and you just say the name of the type (`path/to.Optional`).

For the tests, thinking a bit more and I think what you can do is something like:
- For the generate tests, just check `runtime.Version` and skip the tests if it's too low (feel free to do this in a mildly hacky way, like a list of test case names).
- For the integration tests, I guess it's tricky because they all share one generated file. Maybe make a separate subpackage with a copy of the wiring, genqlient.config, etc., and put `//go:build` lines in each file? Although even that might be awkward because we don't have a great way to put `//go:build` in the `generated.go` itself, and I'm guessing it will break things like `go generate ./...` if we can't parse arbitrary files. And we also don't have a great setup for integration testing different configs. Honestly maybe it's not worth integration testing this right now, you could just leave that out for now, or run it manually and then comment it out so we can hopefully add it back when we're ready to only support 1.18+ (hopefully soon since 1.20 is out).

benjaminjkraft:
Actually, one more thought on integration tests: I'm ok just running all the integration tests only against 1.18+, I don't think there's much we really need to test against older Go versions there. Not sure if that will be enough to make it easy though, so what I said before still applies.

generate/convert.go
DylanRJohnston-FZ: Yeah that's true, I could easily just register the type here. Am I correct in understanding that the return value of `(*generator).ref` is the `packagename.Type` string?

generate/types.go
```diff
@@ -80,6 +81,12 @@ type (
 	// user (perhaps to handle nulls explicitly, or to avoid copying large
 	// structures).
 	goPointerType struct{ Elem goType }
+	// goGenericType represent the Go type ELem, used when requested by the
```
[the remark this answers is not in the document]
DylanRJohnston-FZ: Yeah I planned on updating the docs with examples once I was sure you were on board with the approach. I'll update these comments to explain better too

DylanRJohnston:
@benjaminjkraft I've added a snapshot test and some more documentation. Do you want the the entire integration test to be updated to use the `testutil.Option` or to just leave it for now?

generate/config.go
```diff
@@ -99,6 +100,16 @@ func (c *Config) ValidateAndFillDefaults(baseDir string) error {
 		c.ContextType = "context.Context"
 	}
 
+	if c.Optional != "" && c.Optional != "value" && c.Optional != "pointer" && c.Optional != "generic" {
+		return errorf(nil, "optional must be one of: 'value' (default), 'pointer', or 'generic'")
+	}
+
+	if c.Optional == "generic" && c.OptionalGenericType == "" {
```
benjaminjkraft, line 107: maybe also complain if `c.Optional != "generic" && c.OptionalGenericType != ""`?

go.mod
```diff
@@ -1,15 +1,27 @@
...
+	github.com/davecgh/go-spew v1.1.1 // indirect
+	github.com/gorilla/websocket v1.4.2 // indirect
+	github.com/hashicorp/golang-lru v0.5.0 // indirect
+	github.com/mitchellh/mapstructure v1.2.3 // indirect
+	github.com/pmezard/go-difflib v1.0.0 // indirect
+	golang.org/x/mod v0.6.0-dev.0.20220106191415-9b9b3d81d5e3 // indirect
+	golang.org/x/sys v0.0.0-20220319134239-a9b59b0215f8 // indirect
+	golang.org/x/xerrors v0.0.0-20200804184101-5ec99f83aff1 // indirect
+	gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c // indirect
+)
```
benjaminjkraft, line 27: It seems like this is duplicated now? Did you edit this manually or are the go tools doing something weird?

generate/testdata/snapshots/TestGenerateWithConfig-OptionalGeneric-testdata-queries-generated.go
```diff
@@ -0,0 +1,153 @@
...
+
+// QueryWithSlicesUser includes the requested fields of the GraphQL type User.
+// The GraphQL type's documentation follows.
+//
+// A User is a user!
+type QueryWithSlicesUser struct {
+	Emails                []string                  `json:"emails"`
+	EmailsOrNull          []string                  `json:"emailsOrNull"`
+	EmailsWithNulls       []testutil.Option[string] `json:"emailsWithNulls"`
+	EmailsWithNullsOrNull []testutil.Option[string] `json:"emailsWithNullsOrNull"`
```
benjaminjkraft, line 58: hmm, I wonder if this (which I believe corresponds to a GraphQL `[String]`, i.e. optional list of optional strings) should be `testutil.Option[[]testutil.Option[string]]`. We don't do that for pointers because it's annoying in practice, but if you're using an `Option` type it might be convenient and make it clearer for users to distinguish between `[]` and `null` using `Option`. What do you think?

benjaminjkraft (commented):
Sorry for the delay in reviewing. This seems great!

It's not ideal to have this without integration tests, but I think it's gonna be a decent amount of work to add the support. I've created #261 for this; for now I think it's fine if you just want to test it manually in some other project and that plus snapshot tests should be good enough. Do make sure to at least test it manually as marshaling can go wrong in surprising places!

I suspect that you'll have to either fully drop support for 1.18 (which means removing it from [the list of versions we run tests on](https://github.com/Khan/genqlient/blob/main/.github/workflows/go.yml#L14) as well) or put the additions to `types.go` behind a build flag and add a flag to the test itself. Probably we can just drop 1.18 and that will be easier. (Or if I have time I can do the version changes in a separate PR first, but it's not super likely I will this week or next.) 

Beyond that (and one question for your thoughts below), I think all we need is the changelog entry (or two, if dropping 1.18)!

go.mod
DylanRJohnston-FZ: This is auto generated by `go tidy`. My editor was complaining it needed to be run, but I'm happy to remove ti.

generate/testdata/snapshots/TestGenerateWithConfig-OptionalGeneric-testdata-queries-generated.go
DylanRJohnston-FZ: Generally speaking I prefer to use an empty array instead of option where I can because empty arrays have nice ergonomics, however you may want this in a situation where you're trying to set something to an empty array and it gets interpreted as a missing value. However I've never come across such a situation personally.

benjaminjkraft:
Ok, I pulled the version upgrade into #257, so once that merges you should be able to just merge and avoid dealing with that!

benjaminjkraft:
I went ahead and merged main and added the changelog. Thanks again for writing this -- I'm planning on using it in a project I'm working on!

DylanRJohnston-FZ:
Thanks! Apologies for not pushing this over the line  myself. I just got absolutely swamped at work just after putting this pull request up. Glad to see it in main though!

PR #258, opened 2023-02-22
[draft] feat: Add metrics to genqclient

ra-grover:
Adds prom metrics for GenqClient : 
- A `HistrogramVec` for latency partitioned by operation, code and method.
- `CounterVec` for number of requests  partitioned by operation, code and method
- `CounterVec` for errors (`resp.Errors`)

I have:


ra-grover:
@benjaminjkraft Sorry that I haven't ✅  all the steps in the PR description, I just wanted to gauge the interest from the maintainers if they wanted something like that. I am willing to complete the description/test cases once I have a nod from maintainers.

benjaminjkraft (changes requested):
Thanks for opening this to discuss!

I'm pretty hesitant to pull in one particular metrics collection framework -- many users won't want this as they may be using other tools instead. Unless there's a single framework we can use that will work for everyone (maybe OpenTelemetry is getting close to being that? I don't really know), I'm not sure it makes sense to pull specific frameworks into genqlient. Instead, we should just provide extension points so that users can put in their own tooling.

In theory, my hope is that the existing `Client` interface is sufficient to do that. Would it work for your use cases to wrap the client? You would have your own struct that implements `graphql.Client` and contains a `graphql.Client` returned from genqlient's constructor. Its `MakeRequest` method that would log the on-request metrics you want, then call out to the wrapped client, then log the on-response metrics, then return. If you see issues with that approach, we could look at modifying or extending the existing interface (ideally by adding fields to `Request` or `Response`, since that's not a breaking change, whereas adding methods is).

Potentially we could even look at putting such wrapped clients in a separate package in genqlient, but I think probably best is for that to get prototyped in a separate package, and we can link to that to start with and merge if it becomes popular. That way you don't have to worry about our opinions so much as you iterate :-) .

benjaminjkraft:
Closing this since for housekeeping purposes -- happy to continue to discuss here or in an issue if it's helpful!

PR #262, opened 2023-04-08
Bump minimum Go version to 1.18

### LINKED ISSUE #256: Add Go 1.20 to tests
Go 1.20 is out, so we may as well test against it! See #219 for the last upgrade.

### LINKED ISSUE #257: Drop Go <1.18
This is useful because we can use generics. For example, #252 is gonna be a bit tricky to integration test in part because it would require build tagging a bunch of things. And we can also use generics internally which will probably have at least a little value.

None of this is urgent if people are still using older Go though. @csilvers or @dnerdy, have y'all upgraded to 1.18 yet? Any other users who are still on 1.17 or below?

benjaminjkraft:
This means we can use generics and various other things. I didn't use any of them yet, this is just bumping the numbers. I added tests for 1.20, and fixed one small bug, I think caused by `go/packages` changes therein. And I bumped the golangci-lint version too while I was in the area.

Fixes #256, fixes #257.

I have:


csilvers (approved):
Good by me!

Issue #260, opened 2023-02-27
allow to have omitempty in the settings file for all optional the fields?


liron-navon:
Let's say a graphql query accepts an input object, which has an optional string in it, if I use the input object, and not give a value to that string, it should be committed, that's just how graphql works.
Unfortunately, this library doesn't work that way.

**Is your feature request related to a problem? Please describe.**
I have this input for shopify, the non struct fields are not required

<img width="943" alt="Screenshot 2023-02-27 at 11 29 36" src="https://user-images.githubusercontent.com/18119833/221539762-366b0ae3-df2b-4f9d-bf01-d1e334ba25d9.png">

The generation works well, but if In don't include the optional fields, genqlient turns them into null (when defining value as pointer) or to an empty string (when defining value as value).

They should just be omitted as I don't want to send them, otherwise shopify will fail to use this.

Using the "@genqlient(omitempty: true)" is impossible, because the import is required, even tho every field on it is optional.

**Describe the solution you'd like**
add to the settings.yml file an option "omitempty: boolean" that would just cover all the json fields.

**Describe alternatives you've considered**
Adding manually the "omitempty" directive, but with thousands of fields it's makes it really hard to use, and time consuming and prone to humen errors.

**Additional context**
What I get generated, why would an optional pointer to a string not have omit empty?

<img width="997" alt="Screenshot 2023-02-27 at 11 28 53" src="https://user-images.githubusercontent.com/18119833/221541296-6e11220f-e4e9-4126-84e6-6199fa9b20fe.png">

benjaminjkraft:
Check out the `optional: pointer` setting [in `genqlient.yaml`](https://github.com/Khan/genqlient/blob/main/docs/genqlient.yaml#L100-L112)! If I'm understanding your request correctly that's exactly what it does.

liron-navon:
@benjaminjkraft why did you close it? you don't understand correctly, the "optional: pointer" doesn't omit empty

benjaminjkraft:
Thanks for clarifying. I guess it's valid GraphQL to treat explicit `null` differently. It seems like what you are proposing is a global config to turn on omitempty. That seems reasonable to add, although soon we are going to have to figure out how to make all these related config options not explode the config complexity.

mujagic-nermin:
Looking at the documentation, I'm trying to achieve what it says here with the global omitempty.
[...]

PR #264, opened 2023-04-28
Add `omitempty` for pointer input fields

dharijanto:
When `optional: "pointer"` is configured, `omitempty` is needed for input fields for the generated code to be semantically correct. Without `omitempty` an empty pointer would generate an "empty value" which is incorrect.

Consider the following graphql schema generated by Hasura:

```
input task_insert_input {
  id: Int
}
```

Without this fix, the following golang struct will be generated:
```
  Id: *int `json:"id"`
}
```

When we populate the struct without `Id` being set, it'd serialize to `{ id: 0 }` as oppose to `{}` (i.e. so that the id would be generated by Hasura)

I have:
- [ ] Written a clear PR title and description (above)

### DIFF
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -439,6 +439,11 @@ func (g *generator) convertDefinition(
 				return nil, err
 			}
 
+			if options.GetPointer() || (!field.Type.NonNull && g.Config.Optional == "pointer") {
+				oe := true
+				fieldOptions.Omitempty = &oe
+			}
+
 			goType.Fields[i] = &goStructField{
 				GoName:      goName,
 				GoType:      fieldGoType,


dharijanto:
Created for this: https://github.com/Khan/genqlient/issues/260

generate/convert.go
```diff
@@ -439,6 +439,11 @@ func (g *generator) convertDefinition(
 				return nil, err
 			}
 
+			if options.GetPointer() || (!field.Type.NonNull && g.Config.Optional == "pointer") {
```
dharijanto, line 442: I wonder if this should apply to non-input structures as well..?

generate/convert.go
```diff
@@ -439,6 +439,11 @@ func (g *generator) convertDefinition(
...
+				oe := true
+				fieldOptions.Omitempty = &oe
```
benjaminjkraft, line 444: There's no need to mutate `fieldOptions` -- just compute the correct value and then use it on line 453.

generate/convert.go
benjaminjkraft: omitempty is meaningless for outputs (since it's meaningless on unmarshal)

benjaminjkraft (changes requested):
I think as-is this is in some sense a breaking change, although perhaps anyone who cares about this prefers the new behavior? It also seems confusing to do this only and automatically for `optional: pointer`, which right now is I think otherwise equivalent to `# @genqlient(pointer: true)` on every optional field. Maybe we add a new config `optional: pointer_and_omitempty` and handle both issues that way?

(Edit: Thinking more, doing a new config doesn't strictly handle the second issue -- you could want this for non-`pointer: true` fields. But I guess you can always annotate those individually, and probably you never want it for non-pointer fields globally (since that means `false` gets omitted). So I guess it's still maybe the best compromise.)

BTW, you'll need to update the tests, docs, changelog, etc. -- see the PR message for the checklist.

generate/convert.go
benjaminjkraft: A belated realization: actually the logic here is also wrong: users may need to override the global option locally, if they do need to pass an explicit null. (Sadly in Go, as bad as it is to represent "T or null", it's even worse to represent "T or null or unset", which to be fair is also true of almost any language that's not JavaScript.) That is, if you set this option, then you can do `# @genqlient(omitempty: false)` if you do want explicit null for some query.

So that means you'll need to actually look at the value of `fieldOptions.Omitempty` and then only use this logic if it's null, I think.

partounian:
This would be super awesome. The GraphQL server I'm interacting with has deeply nested inputs and without this feature it breaks.

generate/convert.go
obitech: @benjaminjkraft there doesn't seem to be move movement here on your feedback but I realy really need this feature so I've created https://github.com/Khan/genqlient/pull/287 with your suggestion, in order to get things forward. Feel free to take a look over there :)

benjaminjkraft:
Closing in favor of #308 which uses a config option

Issue #263, opened 2023-04-19
Premarshal struct does not maintain JSON tags (omitempty)


kevinmichaelchen:
**Describe the bug**

When you use custom type bindings, a premarshal struct gets generated.

However, the premarshal struct won't have `omitempty` tags.

I think this is resulting in some buggy behavior when using Hasura.

**To Reproduce**

I tried to do a minimal reproduction here in [this repo](https://github.com/kevinmichaelchen/genqlient-custom-type-omitempty-bug).

Specifically, in [**this comment**](https://github.com/kevinmichaelchen/genqlient-custom-type-omitempty-bug/pull/1/files#r1171607612), I noticed that things work fine when I don't use custom type bindings.

**Expected behavior**

I'm not sure if this is expected behavior or not. Should the premarshal struct not have the same JSON tags as the original struct?

**genqlient version**

Using the latest v0.5.0.

benjaminjkraft:
Thanks for the clear repro! It took me a while to understand the subtle case even looking at all the code so the fact that it was all there was very helpful.

I think you're right that we should just propagate the json tags. It's happening in the first place because the premarshal types are generated in a separate code path ([regular types](https://github.com/Khan/genqlient/blob/main/generate/types.go#L368-L387), [premarshal types](https://github.com/Khan/genqlient/blob/main/generate/marshal.go.tmpl#L31-L39)), which arguably we could change. But anyway that's probably not necessary right now, probably we can just add a check for omitempty in the premarshal code.

I'm hoping to have some time to work on genqlient this weekend and if I do will try to get to this, but if you are excited to, I think this would actually be a good first PR; the code change is probably one line (in the premarshal types linked above; I think the data we need is already available) and the main thing is just adding some unit and integration tests to show that it works.

kevinmichaelchen:
Hey @benjaminjkraft — glad my repro was helpful!

## Me, taking a guess

If I had to take a crack at it, I'd say we want to change

```
    {{.GoName}} {{repeat .GoType.SliceDepth "[]"}}{{ref "encoding/json.RawMessage"}} `json:"{{.JSONName}}"`
    {{else}}
    {{.GoName}} {{.GoType.Reference}} `json:"{{.JSONName}}"`
```

to something like this:

```
    {{.GoName}} {{repeat .GoType.SliceDepth "[]"}}{{ref "encoding/json.RawMessage"}} `json:"{{.JSONName}}{{if .Omitempty -}},omitEmpty{{end}}"`
[...]

PR #267, opened 2023-05-04
fix: premarshal structs get generated with omitempty tag

kevinmichaelchen:
This PR addresses a bug described in https://github.com/Khan/genqlient/issues/263

Basically, whenever a struct involves a custom genqlient binding, a secondary "premarshal" struct gets generated.

The bug was that this "premarshal" struct was not propagating the `omitempty` JSON tags, which was resulting in unexpected behavior.

The fix involved a few changed lines in the Go template, and a few changes in the unit tests.

I have:

### DIFF
--- a/generate/marshal.go.tmpl
+++ b/generate/marshal.go.tmpl
@@ -31,9 +31,9 @@
 type __premarshal{{.GoName}} struct{
     {{range .FlattenedFields -}}
     {{if .NeedsMarshaling -}}
-    {{.GoName}} {{repeat .GoType.SliceDepth "[]"}}{{ref "encoding/json.RawMessage"}} `json:"{{.JSONName}}"`
+    {{.GoName}} {{repeat .GoType.SliceDepth "[]"}}{{ref "encoding/json.RawMessage"}} `json:"{{.JSONName}}{{if .Omitempty -}},omitempty{{end}}"`
     {{else}}
-    {{.GoName}} {{.GoType.Reference}} `json:"{{.JSONName}}"`
+    {{.GoName}} {{.GoType.Reference}} `json:"{{.JSONName}}{{if .Omitempty -}},omitempty{{end}}"`
     {{end}}
     {{end}}
 }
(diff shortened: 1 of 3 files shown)


benjaminjkraft:
Thanks! I'll take a closer look later but a few quick comments now:
- Since this is a bugfix it probably doesn't need any documentation changes, just a brief changelog entry summarizing and linking to the issue.
- If it's not too hard, it would be nice to write an integration test that fails without this change and passes with it! Not sure offhand how hard that will be so if it's too hard it's probably ok to skip.
- Similarly if it's not too hard you could add another unit test case with a schema that looks more like the hasura one. But I'm not sure how interesting that would be, maybe there's no need since existing tests are hitting this case.

kevinmichaelchen:
> a schema that looks more like the hasura one

Here's what I'm thinking for that:
<details>
  <summary>Click me</summary>

```graphql
type Query {
  getPokemon(where: getPokemonBoolExp): [Pokemon!]!
}

input getPokemonBoolExp {
  _and: [getPokemonBoolExp!]
  _not: getPokemonBoolExp
  _or: [getPokemonBoolExp!]
  level: IntComparisonExp
}

input IntComparisonExp {
  _eq: Int
  _gt: Int
  _gte: Int
  _in: [Int!]
  _isNull: Boolean
  _lt: Int
  _lte: Int
  _neq: Int
  _nin: [Int!]
}
```  

</details>

kevinmichaelchen:
I'm actually running into an error when running `UPDATE_SNAPSHOTS=1 go test ./...`

```
generate_test.go:104: no queries found, looked in: testdata/queries/Hasura.graphql (configure this in genqlient.yaml)
```

benjaminjkraft:
Great, thanks! I think `IntComparisonExp` is another limitation, being discussed in #260 (there are a few ways around but none quite convenient for these hasura schemas). Anyway it's good to have the test; I think this weekend I'm going to try to understand what "make genqlient and hasura not fight so much" might mean more generally.

benjaminjkraft (approved):

PR #268, opened 2023-05-05
Update gqlgen to latest

StevenACoffman:
This updates gqlgen to the latest (v0.17.31)

In addition to numerous bug fixes there are a number of new features. [see Changelog](https://github.com/99designs/gqlgen/blob/master/CHANGELOG.md)


benjaminjkraft (approved):
Thanks! gqlgen + testify are super safe since they're dev-only deps. I took a quick look at the gqlparser changelog and everything looks reasonable (I'm very excited about vektah/gqlparser#234, I was just complaining to @MiguelCastillo about this!). The changelog for `x/tools` is way too large to look at but our use is not too deep so fingers crossed. Let's do it!

PR #282, opened 2023-07-16
Update gqlparser and gqlgen dependencies

StevenACoffman:
This updates both gqlparser and gqlgen dependencies. This was prompted by an error report in gqlparser.

gqlparser also preserves comments now, so some test cases might need tweaking here.

Per @tduong2049 in https://github.com/vektah/gqlparser/issues/269 

> Prior to [v2.5.2](https://github.com/vektah/gqlparser/releases/tag/v2.5.2), gqlparser parses the supergraph and returns no errors. 
> Upon upgrading to v2.5.2+, I get the following error when running `go run github.com/Khan/genqlient`:
> `invalid schema: Argument extension for directive join__type cannot be null.`

> Possibly related to #258?

### What did you expect?
No parsing errors when running: `go run github.com/Khan/genqlient` on a supergraph. 
However, a validation error is returned when parsing `@join__type` directives.

### Minimal graphql.schema and models to reproduce
```
type AircraftManufacturer
  @join__type(graph: AIRCRAFT, key: "uuid")
{
  uuid: ID!
  name: String!
}
```


csilvers (approved):
Seems safe enough!

StevenACoffman:
@benjaminjkraft Hmmm, I took a quick glance, and it's not obvious to me how to update the tests here.
The tests now complain that the ast.Dump contains the comments:
```
              Comment: "# @genqlient\n"
```

csilvers:
> @benjaminjkraft Hmmm, I took a quick glance, and it's not obvious to me how to update the test expectations here. The tests now complain that the ast.Dump contains the comments:
> 
> ```
>               Comment: "# @genqlient\n"
> ```

Did you run `UPDATE_SNAPSHOTS=1 go test ./...`?  It's not clear to me which tests are failing for you.  Did you look at docs/CONTRIBUTING.md?

StevenACoffman:
@csilvers Yeah, I updated the snapshots, but that doesn't update these particular expectations, which are reused for other purposes. 
I pushed a commit that just removes the comment lines from the ast.Dumps for the expectations, which is hacky, but clarifies where the problem is. I'm minded to merge this as is and allow someone else to more elegantly solve it in a follow-up.

benjaminjkraft:
The test fix seems fine, in theory it should probably remove comments from both sides of the assertion but we can fix that if/when it matters.

But I am a little worried about the changes where we now copy comments, including the `# @genqlient`, into the generated code (and thereby into the request to the server). It's both extra bytes on the wire (probably ok in practice, it's not like we are bothering to minify) and would mean that if your genqlient.yaml is loosely configured you could get circularity (if it tries to extract queries from the generated file). Let me take a look at fixing that.

benjaminjkraft:
Turns out we need vektah/gqlparser#271 to make it easy to remove the comments.

PR #284, opened 2023-07-17
Update gqlparser to omit comments

StevenACoffman:
See https://github.com/vektah/gqlparser/pull/271

This updates gqlparser dependency to v2.5.8, which defaults to not including comments.
This PR will update the snapshots and generated files to match that. The AST will still include the comments in the dump, so the tweak to that one test in parse_test.go is still necessary, so that remains in place.


benjaminjkraft (approved):
Thanks! [Diff](https://github.com/Khan/genqlient/compare/main%5E...update_gqlparser_v2_5_8) from the two PRs combined looks great.

PR #285, opened 2023-07-18
feat: support binding to generated package

### LINKED ISSUE #283: Binding to a type in the same package as generated code produces invalid output
**Describe the bug**
Binding to a type in the same package as the generated file is not detected and results in an import loop as the package is import itself.

**To Reproduce**
Output the generated file in the same location as a bound type.
```yaml
generated: db/generated.go
bindings:
  FilterType:
    type: github.com/mathewbyrne/project/db.Filter
```

**Expected behavior**
The generated output should not have any import statement, as the type is in the same package. The type should be used directly without reference to it's package.

**genqlient version**
0.6

**Additional context**
Working around this is fairly simple by placing the generated output in a sub-package, but this still seems like a bug that could be resolved.

costela:
This allows one to generate code in an existing package without causing import loops.

It is now possible to use something like the following (assuming we're in module `github.com/foo/bar`):
```yaml
generated: generated/genqlient.go
package_bindings:
  - github.com/foo/bar/generated
```
or (shorthand variant of the example above)
```yaml
generated: generated/genqlient.go
package_bindings:
  - .
```
or
```yaml
generated: generated/genqlient.go
bindings:
  Bar:
    type: github.com/foo/bar/generated.Bar
```

The last example assumes some file in `generated/` defines the type `Bar`.

Closes #283


generate/generate_test.go
```diff
@@ -21,24 +21,41 @@ const (
 // buildGoFile returns an error if the given Go code is not valid.
 //
 // namePrefix is used for the temp-file, and is just for debugging.
-func buildGoFile(namePrefix string, content []byte) error {
+func buildGoFile(namePrefix string, content []byte, extraFiles ...string) error {
```
benjaminjkraft, line 24: nice!

docs/genqlient.yaml
```diff
@@ -224,6 +224,9 @@ bindings:
 # to the bindings map, above, for each exported type in the package.  Multiple
 # packages may be specified, and later ones take precedence over earlier ones.
 # Explicit entries in bindings take precedence over all package bindings.
+#
+# Using "." is a shorthand for the package containing the generated file.
+# Both allow the generated code to live in the same package as its types.
```
benjaminjkraft, line 229: FWIW, I would have some concerns about actually using the generated package here. I worry it would end up being circular in ways that are hard to understand -- in that now genqlient depends on genqlient-generated types. I don't think we need to ban it, but maybe we should say in the comment here that it's not a good idea? Or does it actually work fine?

Also, whatever we settle on for the syntax, can you add it to the docs for `bindings` above too?

benjaminjkraft (changes requested):
Cool, thanks for figuring this out! I'm very happy with how little code it turned out to be, and the tests are very convincing that it works.

One question I have is whether we should allow any valid relative path in the `genqlient.yaml`, i.e. you could say `./mypkg/subpkg` if that's what you want. That seems like it could be quite convenient rather than writing out the name of the current package a bunch of times. In that case we'd probably want to say `.` means the directory with the `genqlient.yaml`. Although, maybe it's better to say you have to write it out, like you do in Go import blocks. In fact I think these changes would solve the problem without allowing special `.` syntax at all -- you would just do what you did in the first two test cases. I think I'd prefer to either fully support the usual, or stick to what we have, since config syntax is always confusing and can be more annoying to change later.

Beyond that I think my questions are about whether running this in the real world could hit weird cases we don't really see in a test. I put some comments inline. But probably if we can't figure those out we can just see if anyone complains.

docs/genqlient.yaml
costela: > now genqlient depends on genqlient-generated types

Yes, a warning might be appropriate, but it should work as long as the user avoids name conflicts.
I've added a "double generate" test to try to cover this case and made it more explicit in the comment.

> can you add it to the docs for bindings above too

The shorthand is intentionally only supported in the "whole package" mode. Otherwise one would end up with `..ID` or some other weird syntax to reference the types inside the package. Just didn't feel like it was worth it. WDYT?

generate/config.go
```diff
@@ -153,6 +156,23 @@ func (c *Config) ValidateAndFillDefaults(baseDir string) error {
...
+	if err != nil {
+		return fmt.Errorf("finding relative path to generated package: %w", err)
+	}
+	pkgs, err := packages.Load(&packages.Config{
+		Dir:  c.baseDir,
+		Mode: packages.NeedName | packages.NeedModule,
+	}, "./"+filepath.Dir(genPkgPath))
+	if err != nil {
+		return fmt.Errorf("loading generated package: %w", err)
+	}
```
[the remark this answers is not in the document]
costela: > would this break things if the directory where you're putting the output is empty

Yes, it would break, but that would be the same as trying to bind to an empty package, right?

> does this require that baseDir is a valid Go package?

It shouldn't: it's just used as a base for the package search.

generate/config.go
```diff
@@ -203,6 +223,7 @@ func (c *Config) ValidateAndFillDefaults(baseDir string) error {
 
 			mode := packages.NeedDeps | packages.NeedTypes
 			pkgs, err := packages.Load(&packages.Config{
+				Dir:  c.baseDir,
```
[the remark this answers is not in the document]
costela: yes, exactly. This is just needed to support the "relative import" hack we use above, which isn't usual, but is supported.

costela (commented):
Thanks for the feedback and sorry for the late reply! Vacation got in the way!

PTAL again! :pray:

benjaminjkraft:
Closing in favor of #316 which makes this as a narrower change to support existing syntax. If you're interested, you could revisit adding the shorthand (`package_bindings: .` etc.); I think doing that separately will make some of the still-questions about what syntax we should support clearer.

Issue #272, opened 2023-05-06
Document and improve support for Hasura schemas


benjaminjkraft:
We seem to have a lot of users using genqlient to query a [Hasura](https://hasura.io/docs/latest/index/)-generated schema, and they aren't having a great time of it; a lot of bugs and feature requests have come up over time. It would be good for us to take the time to do something a little more holistic, to see what support should look like. We can then document that, fix any gaps, etc.

## The Problems

From reading all the issues mentioning Hasura, it seems like the main thing is the [query syntax](https://hasura.io/docs/latest/queries/postgres/query-filters/#the-where-argument) which has several interesting properties (discussed in #149 and #260):
- it generates complex and often self-referential types
- you usually only want to set a few fields of the very large input objects
- it distinguishes implicit and explicit null ([example](https://github.com/hasura/graphql-engine/blob/master/CHANGELOG.md#support-for-null-values-in-boolean-expressions))
These conflict badly with genqlient's defaults, and arguably with mapping the types to Go itself. (Some of these issues also apply to [mutation inputs](https://hasura.io/docs/latest/mutations/postgres/insert/#insert-a-single-object), but those are a bit closer to normal mutation inputs, and you probably use them in fewer places.)

There are also some other specific problems, which are more orthogonal but any documentation should mention too:
- hasura type names (both input and output) are non-idiomatic: #206 

From #206, the lunasec folks have a decently large [example schema](https://github.com/lunasec-io/lunasec/blob/master/lunatrace/gogen/schema.graphql) as an example. There's also a [schema](https://github.com/Khan/genqlient/files/7437100/schema.graphql.zip) in #149.

## What should we generate?

Some thinking out loud.

First of all, one theory -- already very possible -- is to say you just want to [write your input object in the query](https://github.com/Khan/genqlient/issues/260#issuecomment-1533975590). This is obviously less flexible, which in the common case is arguably good but doesn't scale well if you want to construct your queries with arbitrary Go logic, and doesn't handle certain cases like arbitrary-length lists of objects (although short of jsonb and arbitrary-length boolean it's not clear those really exist -- this is SQL after all).

[...]

PR #287, opened 2023-07-27
feat: always set omitempty on pointer fields

obitech:
When optional: "pointer" is configured, omitempty is needed for input fields for the generated code to be semantically correct. Without omitempty an empty pointer would generate an "empty value" which is incorrect.

Consider the following graphql schema generated by Hasura:

```
input task_insert_input {
  id: Int
}
```

Without this fix, the following golang struct will be generated:

```
  Id: *int `json:"id"`
}
```

* Ref: https://github.com/Khan/genqlient/issues/260
* Ref: https://github.com/Khan/genqlient/pull/264

I have:

### DIFF
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -14,6 +14,8 @@ When releasing a new version:
 
 ### Bug fixes:
 
+- Using `pointer: true` will generate `omitempty` tags to all pointer input types.
+
 -->
 
 ## next
--- a/generate/convert.go
+++ b/generate/convert.go
@@ -453,7 +453,7 @@ func (g *generator) convertDefinition(
 			// We always want to set omitempty if the field is a pointer, unless
 			// the user explicitly wants the value to be sent as null.
 			omitEmpty := fieldOptions.GetOmitempty()
-			if options.GetPointer() && (!field.Type.NonNull && g.Config.Optional == "pointer") && fieldOptions.Omitempty == nil {
+			if options.GetPointer() || (!field.Type.NonNull && g.Config.Optional == "pointer") && fieldOptions.Omitempty == nil {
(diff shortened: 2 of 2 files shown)


benjaminjkraft (changes requested):
Thanks for picking this up!

I think my comment in the original PR that we'll need this flagged somehow is still relevant:

> I think as-is this is in some sense a breaking change, although perhaps anyone who cares about this prefers the new behavior? It also seems confusing to do this only and automatically for `optional: pointer`, which right now is I think otherwise equivalent to `# @genqlient(pointer: true)` on every optional field. Maybe we add a new config `optional: pointer_and_omitempty` and handle both issues that way?
> 
> (Edit: Thinking more, doing a new config doesn't strictly handle the second issue -- you could want this for non-`pointer: true` fields. But I guess you can always annotate those individually, and probably you never want it for non-pointer fields globally (since that means `false` gets omitted). So I guess it's still maybe the best compromise.)

I'm not sure if it makes more sense to do a completely separate option, or a separate value for `optional`, or what, but we'll need one or the other! Eventually maybe we change the default here but I'd like to see more of how `optional: generic` (and generally Go style around how to represent `Optional[T]`) develops, if nothing else, so I think for now it's best to avoid that. (Plus, really, what we need is a Hasura template -- at which point it doesn't matter what the defaults are. A reminder to comment on #272 if you have any thoughts!)

And it still needs some tests and documentation for the new behavior :-) .

justonia:
What's the status on this? I'm working with the Linear API and it's broken if `omitempty` isn't supported.

staugaard:
I'm also waiting for this feature.

benjaminjkraft:
Closing in favor of #308 which uses a config option

PR #289, opened 2023-07-28
feat: allow imports from same generated package

avinassh:
This is a small change, but it lets you import marshaler and unmarshalers from the same generated package. 

Why?
   I'd like to keep all the graphql related code into one. Following the conventions of the `genqlient/example`, I would like to keep a custom marshaler in the same directory as `example`. For e.g., `example/helper.go` could contain:

```go
// helper.go
func UnMarshalDateTime(b []byte, v *time.Time) error {
	...
	return nil
}
```

 But following won't work:

```yaml
...
generated: generated.go
package: main
bindings:
  DateTime:
    type: time.Time
    unmarshaler: "https://github.com/Khan/genqlient/main.UnMarshalDateTime" // since package name is main
    # unmarshaler: "https://github.com/Khan/genqlient/example.UnMarshalDateTime" // nor this, if one is using a custom package
```

So, instead I would like to add a new feature where one can import from the same package:

```yaml
...
generated: generated.go
package: main
bindings:
  DateTime:
    type: time.Time
    unmarshaler: "main.UnMarshalDateTime"
```

What do you think? If you think this is good, then I will add tests and update the documentation

I have:

### DIFF
--- a/generate/imports.go
+++ b/generate/imports.go
@@ -100,6 +100,11 @@ func (g *generator) ref(fullyQualifiedName string) (qualifiedName string, err er
 	pkgPath := nameToImport[:i]
 	localName := nameToImport[i+1:]
 	alias, ok := g.imports[pkgPath]
+
+	if pkgPath == g.Config.Package {
+		return localName, nil
+	}
+
 	if !ok {
 		if g.importsLocked {
 			return "", errorf(nil,


benjaminjkraft (changes requested):
Thanks for the pull request!

Did you see #285? I think you're both trying to solve the same issue. I like the narrower approach here -- as I said on that PR I think if we introduce new syntax we need to make sure we get it right whereas you sidestep that. But I don't think what you have quite works as-is, because the [`package`](https://github.com/Khan/genqlient/blob/main/docs/genqlient.yaml#L32-L34) option is the package name, not the full package path.

Relatedly, a test would help show whether it works! And is definitely necessary -- along with a changelog entry -- to merge.

benjaminjkraft:
Closing in favor of #316,

avinassh:
Hey @benjaminjkraft I missed the previous notification. I checked #316 and I think its in the right direction 👍

Issue #279, opened 2023-06-22
codegen error: 'data' redeclared in this block


xmlking:
I am using **Hasura** scheme. the generated code is using a redeclared variable which is causing `'data' redeclared in this block` error. 

I have to mutually edit  `UpsertDeviceResponse` from `data` to `data2` 
is there any workaround?

### manual fix
```go
func UpsertDevice(
	ctx context.Context,
	client graphql.Client,
	data Devices_insert_input,
) (*UpsertDeviceResponse, error) {
	req := &graphql.Request{
		OpName: "UpsertDevice",
		Query:  UpsertDevice_Operation,
		Variables: &__UpsertDeviceInput{
			Data: data,
		},
	}
	var err error

	var data2 UpsertDeviceResponse
	resp := &graphql.Response{Data: &data}

	err = client.MakeRequest(
		ctx,
		req,
		resp,
	)

	return &data2, err
}
```

### query
```gql
mutation UpsertDevice($data: devices_insert_input!) {
    insert_devices_one(object: $data, on_conflict: {update_columns: [ip, description, tags, annotations], constraint: devices_display_name_organization_key, where: {deletedAt: {_is_null: true}}}) {
        annotations
        ip
        id
        displayName
        ...
    }
}
```
'data' redeclared in this block

<img width="743" alt="image" src="https://github.com/Khan/genqlient/assets/627951/18a493a5-6596-42b5-90ee-29e59139eb57">

<img width="1071" alt="image" src="https://github.com/Khan/genqlient/assets/627951/5e868f44-fb83-4213-8bc2-aeb2e3150cc0">

benjaminjkraft:
Common cases of this are fixed in #291. Leaving this open for a general fix.

PR #291, opened 2023-07-31
bugfix: local variable 'data' colliding with argument name

zholti:
This PR makes [THIS ISSUE](https://github.com/Khan/genqlient/issues/279) a lot less likely to occur. 
The local name "data" in generated functions was colliding with arguments called "data". 

Note that no explicit name collision check was added, but "data" inside the function was just renamed to "data_" to make it less likely to collide with arguments.

I have:


generate/testdata/snapshots/TestGenerate-QueryArgCalledData.graphql-QueryArgCalledData.graphql.json
benjaminjkraft, line 1: I think these snapshots correspond to a query that's not checked in -- did you forget to `git add` the `QueryArgCalledData.graphql`? (We should, but don't, have a test for orphaned snapshots...)

benjaminjkraft (changes requested):
Thanks -- this seems like a fine partial solution. Eventually we should probably do some fancy name-mangling but `data` is probably the biggest issue.

Could you add a changelog entry, as well as the missing snapshot I mention below?

zholti:
> @benjaminjkraft should weigh in too, bit it seems like a good idea to me! Can you add the underscore to all vars, not just`data`?

Did that for the local "err" as well. The collision with query/mutation args affects only this template as far as I can see. Which other vars in which templates can be affected?

benjaminjkraft (approved):
As far as I can tell this is the only template that uses user-defined variable names. To be complete we should do the same for `req` and `resp`. If you get a chance to do that that would be great. (In theory we should also do `ctx` and `client`, but since those are user-visible, I'd prefer if we can change them only if needed. And they're probably less likely names to conflict with, since context is a Go thing and how would you pass a client over GraphQL.)

csilvers:
I could see you passing in a var called `client` to say which client to use. But I agree we don't have to handle it now if it's user facing

zholti:
> I could see you passing in a var called `client` to say which client to use. But I agree we don't have to handle it now if it's user facing

Passing a client is already there, no? Depending on Config.ClientGetter, either the one passed as arg or the local one will be constructed.

But you do have a point - no matter if the client is passed or constructed, an arg called "client" will collide with it in both cases.

zholti:
> As far as I can tell this is the only template that uses user-defined variable names. To be complete we should do the same for `req` and `resp`. If you get a chance to do that that would be great. (In theory we should also do `ctx` and `client`, but since those are user-visible, I'd prefer if we can change them only if needed. And they're probably less likely names to conflict with, since context is a Go thing and how would you pass a client over GraphQL.)

In my latest commit, I renamed client to client_, since I can imagine that there might be schemas with fields taking such an arg. Should I also do the same treatment on ctx? This is less likely to collide with argument names, but still...

benjaminjkraft:
If we're gonna do `client_`, we may as well do the same for `ctx_`!

Issue #199, opened 2022-05-24
Support for GraphQL Subscriptions


pixelsquared:
**Is your feature request related to a problem? Please describe.**
I am working with an API that uses subscriptions and I would like to use them instead of polling.

**Describe the solution you'd like**
I was thinking a function that returns a channel for the subscription?

**Describe alternatives you've considered**
The alternative is polling

benjaminjkraft:
Thanks for raising this!

I took a quick look at the [Hasura fork](https://github.com/hasura/go-graphql-client/blob/master/subscription.go) of the shurcooL client, which seems to be the only Go client that supports subscriptions, as well as the [transport spec](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) and [client implementation](https://github.com/enisdenjo/graphql-ws/blob/master/src/client.ts) Apollo uses/recommends.  It looks like it will be a decent amount of work; I think it should be very possible to fit into the genqlient model and some parts may be simpler but we'll definitely need someone who's excited to implement it to make it happen.

I think much of the work will simply be in writing the client code, which should be similar to what other clients do.  We'll also need to figure out what it actually looks like to open/close a subscription with genqlient; presumably you pass a callback (with types we generate?) to the entrypoint function to open a subscription but there are some details to work out for what to do on error and how to close.  My guess is wiring up the types there won't be too hard; it should just be like a query in terms of the types we generate, and it's the control flow that's very different.

Let me know if you are interested in taking a stab at that!  (Or if anyone else reading this is.)  I'm happy to provide any guidance you need for how to integrate with genqlient, or schedule a call to talk about it.

pixelsquared:
I am interested in working on this. I am going to do some reading from the links you gave and look in to how genqlient works a bit more.

gedw99:
what transport ? http polling, websockets, sse ???

lander2k2:
@pixelsquared Have you begun working on this?  I'm also interested in adding subscription support but don't want to if you already have something in the works...

pixelsquared:
@lander2k2 I have been able to work on this yet. So go ahead.

gedw99:
might be interested in how they used graphql and golang: https://github.com/borderlesshq/graphrpc

lander2k2:
Have had to de-prioritize this :(    May return to it later...

jjauzion:
Hello all, is anyone working on this feature ?

jjauzion:
[...]

PR #294, opened 2023-09-04
Subscription support

matthieu4294967296moineau:
**Added [support for graphQL subscriptions](https://github.com/Khan/genqlient/issues/199).**
- Using `graphql.NewClientUsingWebSocket()`, the returned `graphql.WebSocketClient` will be able to subscribe to graphQL endpoints.
- Implementation does not depend on a specific websocket package, but it is similar to [github.com/gorilla/websocket](https://github.com/gorilla/websocket) (this package is also used to create a client in the tests).

I have:


StevenACoffman:
Hi! In addition to your CEO, you also need to sign the CLA as yourself, since you are the one contributing the code. Thanks!

matthieu4294967296moineau:
> Hi! In addition to your CEO, you also need to sign the CLA as yourself, since you are the one contributing the code. Thanks!

Done !

graphql/client.go
```diff
@@ -38,10 +38,38 @@ type Client interface {
...
+	//
+	// errChan is a channel on which are sent the errors of webSocket
+	// communication.
+	//
+	// err is any error that occurs when setting up the webSocket connection.
+	DialWebSocket(
+		ctx context.Context,
+		req *Request,
+		respChan chan json.RawMessage,
+	) (errChan chan error, err error)
```
benjaminjkraft, line 63: Why is `respChan` a parameter but `errChan` a return? Is this a standard thing?

Also, I definitely regretted not initially making `Response` a struct -- thoughts on changing this to `resp *WebsocketResponse` which would have `respChan` (and maybe `errChan`) as fields so we can add more later?

internal/integration/roundtrip.go
```diff
@@ -126,3 +136,24 @@ func newRoundtripGetClient(t *testing.T, endpoint string) graphql.Client {
 		t:         t,
 	}
 }
+
+type MyDialer struct {
```
benjaminjkraft, line 140: Thinking out loud: would it be useful to test against multiple clients here? e.g. `x/net/websocket` or `nhooyr.io/websocket` or whatever is up there on grank.io. Not a big deal if there are issues but would help prove out if the interface can actually support different clients.

(If `x/net/websocket` works out, we could also use that in the example (with a comment suggesting others) and keep the main `go.mod` a bit smaller.)

exampleSubscription/README.md
benjaminjkraft, line 1: can you add this to `example_test.go` to make sure it stays up to date?

graphql/websocket.go
```diff
@@ -0,0 +1,186 @@
...
+	webSocketTypeConnAck    = "connection_ack"
+	webSocketTypeSubscribe  = "subscribe"
+	webSocketTypeNext       = "next"
+	webSocketTypeError      = "error"
+	webSocketTypeComplete   = "complete"
+	websocketConnAckTimeOut = time.Second * 30
+)
+
+// Close codes defined in RFC 6455, section 11.7.
+const (
```
benjaminjkraft, line 23: can these, and below, be private? since hopefully the caller gets them from their client anyway.

graphql/websocket.go
```diff
@@ -0,0 +1,186 @@
...
+
+const (
+	webSocketMethod         = "websocket"
+	webSocketTypeConnInit   = "connection_init"
+	webSocketTypeConnAck    = "connection_ack"
+	webSocketTypeSubscribe  = "subscribe"
+	webSocketTypeNext       = "next"
+	webSocketTypeError      = "error"
+	webSocketTypeComplete   = "complete"
+	websocketConnAckTimeOut = time.Second * 30
```
benjaminjkraft, line 19: silly question, are these GraphQL things or websocket things or what? is there some spec or documentation we can link to that lists them? (are they things a client might need to configure in the future?)

exampleSubscription/main.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	dialer := websocket.DefaultDialer
+	headers := http.Header{}
+	headers.Add("Authorization", "bearer "+key)
+
+	graphqlClient := graphql.NewClientUsingWebSocket(
+		"wss://api.github.com/graphql",
+		&MyDialer{Dialer: dialer},
+		headers,
```
benjaminjkraft, line 45: Wait, does the real GitHub API even have subscriptions? If not, I don't think this will work, right? We could either include a gqlgen server with the example, or just skip the full example in the repo entirely and provide the relevant snippets in the documentation. (Eventually maybe we should have many examples each with their own gqlgen server.)

benjaminjkraft (changes requested):
Wow, thanks so much for adding this! On the whole it looks great, I think almost everything is integrated in the right places and in particular I'm very happy with how minimal the changes to `generate` are, which I hope will keep this very maintainable.

I made a bunch of small comments and things to tweak, including some probably-foolish questions since I haven't worked with GraphQL subscriptions much. I didn't really get a chance to review the core websocket logic yet, but I figured I'd post everything else now. If anyone else has more experience with websockets a review of those parts would be very helpful!

One other thing is I think this will need some documentation, for now the best place is probably to add something in `FAQ.md` as well as of course under features in the changelog!

generate/operation.go.tmpl
```diff
@@ -1,7 +1,14 @@
-// The query or mutation executed by {{.Name}}.
+// The query, mutation or subscription executed by {{.Name}}.
```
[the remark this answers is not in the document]
matthieu4294967296moineau: done

graphql/websocket.go
matthieu4294967296moineau:
```
        webSocketMethod         = "websocket"
```
this is just an identifier I chose to identify the websocket method from the `http.MethodGet` and `http.MethodPost`

```
	webSocketTypeConnInit   = "connection_init"
	webSocketTypeConnAck    = "connection_ack"
	webSocketTypeSubscribe  = "subscribe"
	webSocketTypeNext       = "next"
	webSocketTypeError      = "error"
	webSocketTypeComplete   = "complete"
```
these are graphQL constants, they are defined [here](https://github.com/99designs/gqlgen/blob/fceb33111bef2bbbf14468d78888098ac8968351/graphql/handler/transport/websocket_graphql_transport_ws.go#L14)

```
        websocketConnAckTimeOut = time.Second * 30
```
this is arbitrary, it could be added to a webSocket client configuration

exampleSubscription/main.go
matthieu4294967296moineau: I added the example thinking we could add a subscription on the server side.
But for now I will just replace it with documentation.

exampleSubscription/README.md
matthieu4294967296moineau: I will replace the example with documentation, so no need for this now.

.golangci.yml
```diff
@@ -49,6 +49,7 @@ linters-settings:
       - golang.org/x/tools
       - gopkg.in/yaml.v2
       - github.com/alexflint/go-arg
+      - github.com/gorilla/websocket
```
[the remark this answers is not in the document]
matthieu4294967296moineau: example will be removed for now and I add a doc instead

generate/operation.go.tmpl
```diff
@@ -36,14 +43,60 @@ func {{.Name}}(
         return nil, {{if .Config.Extensions -}}nil,{{end -}} err_
     }
     {{end}}
-    var data_ {{.ResponseName}}
-    resp_ := &graphql.Response{Data: &data_}
+    {{if eq .Type "subscription"}}
+    dataChan_ = make(chan {{.Name}}WsResponse, 1)
+    respChan_ := make(chan json.RawMessage, 1)
```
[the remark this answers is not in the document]
matthieu4294967296moineau: I used [99designs code](https://github.com/99designs/gqlgen/blob/fceb33111bef2bbbf14468d78888098ac8968351/graphql/handler/transport/websocket.go#L128) as a model for the buffered channel.
I am not sure whether this is standard or if it is their implementation.

generate/operation.go.tmpl
```diff
@@ -15,7 +22,7 @@ func {{.Name}}(
     {{.GraphQLName}} {{.GoType.Reference}},
     {{end -}}
     {{end -}}
-) (*{{.ResponseName}}, {{if .Config.Extensions -}}map[string]interface{},{{end}} error) {
+) ({{if eq .Type "subscription"}}dataChan_ chan {{.Name}}WsResponse, errChan_ chan error,{{else}}data_ *{{.ResponseName}}, {{if .Config.Extensions -}}ext_ map[string]interface{},{{end}}{{end}} err error) {
```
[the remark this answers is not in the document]
matthieu4294967296moineau: done

generate/operation.go.tmpl
```diff
@@ -1,7 +1,14 @@
...
 const {{.Name}}_Operation = `{{$.Body}}`
 
 {{.Doc}}
+{{if eq .Type "subscription"}}
+{{if eq .Doc ""}}
+// {{.Name}}
+{{end -}}
+//
+// To close the connection, use [graphql.Client.CloseWebSocket()]
+{{end -}}
```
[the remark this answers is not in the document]
matthieu4294967296moineau: done

graphql/websocket.go
matthieu4294967296moineau: done

generate/operation.go.tmpl
```diff
@@ -36,14 +43,60 @@ func {{.Name}}(
...
+}
+
+{{if eq .Type "subscription"}}
+type {{.Name}}WsResponse struct {
+	Data       *{{.ResponseName}}     `json:"data"`
+	Extensions map[string]interface{} `json:"extensions,omitempty"`
+	Errors     error                  `json:"errors"`
+}
+
+func {{.Name}}ForwardData(dataChan_ chan {{.Name}}WsResponse, respChan_ chan json.RawMessage, errChan_ chan error) {
```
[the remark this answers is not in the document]
matthieu4294967296moineau: yes the webSocket client sends errors on the errChan_

graphql/client.go
matthieu4294967296moineau: I did it this way because the `respChan`'s type depends on the type of the returned object, while the `errChan` always sends an error, so it could be added to the `webSocketClient` struct.
But if it is disturbing, I could bring back the `errChan` as a parameter, just like for the `respChan`.

internal/integration/roundtrip.go
matthieu4294967296moineau: I tested with `x/net/websocket` and it works, I will add to the doc both ways: with `x/net/websocket` and with `gorilla/websocket`.
The package `gorilla/websocket` was in the `go.mod` because it was the client I used for the tests.
But now I can change and use `x/net/websocket` in the tests if you prefer, even if it adds `golang.org/x/net` as direct dependency, and `golang.org/x/sys` as indirect.

graphql/client.go
```diff
@@ -71,23 +102,41 @@ func NewClient(endpoint string, httpClient Doer) Client {
...
 func NewClientUsingGet(endpoint string, httpClient Doer) Client {
-	return newClient(endpoint, httpClient, http.MethodGet)
+	return newClient(endpoint, httpClient, http.MethodGet, nil)
+}
+
+// NewClientUsingWebSocket returns a [Client] which makes subscription requests
+// to the given endpoint using webSocket.
+//
+// The client does not support queries nor mutations, and will return an error
+// if passed a request that attempts one.
```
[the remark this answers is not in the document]
matthieu4294967296moineau: I agree with you, I made `WebSocketClient` a separate interface

matthieu4294967296moineau (commented):
I answered to all the comments so far.
Either I directly pushed the fix/update, or I answered the comment when not sure what decision should be made.

What do you think of the way errors are returned ?
It might be a bit confusing to have 3 possibilities to receive errors:
- respChan: where api errors are returned in the response
- errChan: where webSocket errors are sent
- err: initialization error

generate/operation.go.tmpl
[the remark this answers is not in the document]
benjaminjkraft: Hmm, poking at gorilla/websocket I think it doesn't check the ctx. That means we need to -- something like the suggestions in this issue: https://github.com/gorilla/websocket/issues/733

graphql/client.go
```diff
@@ -177,7 +249,73 @@ func (c *client) MakeRequest(ctx context.Context, req *Request, resp *Response)
...
+		w.conn.Close()
+		return err
+	}
+	err = w.waitForConnAck()
+	if err != nil {
+		w.conn.Close()
+		return err
+	}
+
+	go w.listenWebSocket(respChan)
```
benjaminjkraft, line 301: So I guess we end up with two goroutines per subscription, one in the `graphql.Client` to wait on the websocket client and forward to `respChan`, and then one from the generated code that waits on `respChan` and forwards to `dataChan` and `errChan`. Can we avoid that? (Or at least, can we set up the interface so it's possible to avoid in the future, even if it's not worth the effort now?)

benjaminjkraft (changes requested):
Thanks for all the replies! I am learning so much about websockets and graphql-ws, and now I have a bunch more questions (sorry!).

The most important question, I realized -- shouldn't the client open a single websocket for all subscriptions rather than one per subscription? I don't know how many subscriptions the average client will have but whenever it's more than one I bet the server prefers that. It looks like that's what `graphql-ws` (used by `apollo-client` and thus the defacto standard) [does](https://github.com/enisdenjo/graphql-ws/blob/master/src/client.ts#L625).

Two ways to do this, I guess: we could do it at init (or as a separate method the caller is supposed to call first), or do it lazily. Probably lazily is less surprising? (Again that's what `graphql-ws` does.) Either way, we'll need to keep some plumbing to direct the right messages to the right channels based on their IDs. And we'll probably need separate close-subscription and close-client methods. (Or do we need close-client? I feel like http clients don't tend to come with those, even when they keep a connection pool. (Ok, looks like `http.Transport` has a `CloseIdleConnections`, which is semi-equivalent. Maybe a good idea to be safe.)

Or, we don't necessarily need to implement that now. But at a bare minimum I would like to make sure all the API we're adding has what we need. Maybe it mostly already does, but at least we should rename it to `Subscribe` since it won't necessarily dial, and figure out the semantics of the close methods.

graphql/websocket.go
```diff
@@ -0,0 +1,155 @@
...
+
+type webSocketReceiveMessage struct {
+	Type    string          `json:"type"`
+	Payload json.RawMessage `json:"payload"`
+}
+
+func (w *webSocketClient) sendInit() error {
+	connInit := webSocketSendMessage{
+		Type: webSocketTypeConnInit,
+	}
```
simluk, line 52: So in our case we need to provide bearer token in `connection_init` message:
```
{
  "type": "connection_init",
  "payload": {
    "Authorization": "Bearer token"
  }
}
```
Would it be possible to provide extension point for this ?

graphql/client.go
```diff
@@ -83,6 +123,24 @@ func NewClientUsingGet(endpoint string, httpClient Doer) Client {
...
+// NewClientUsingWebSocket returns a [WebSocketClient] which makes subscription requests
+// to the given endpoint using webSocket.
+//
+// The client does not support queries nor mutations, and will return an error
+// if passed a request that attempts one.
+func NewClientUsingWebSocket(endpoint string, wsDialer Dialer, headers http.Header) WebSocketClient {
+	if headers == nil {
+		headers = http.Header{}
+	}
+	headers.Add("Sec-WebSocket-Protocol", "graphql-transport-ws")
```
simluk, line 135: Do you need to force `graphql-transport-ws` sub-protocol? We are using `graphql-ws` in our case. Had to override this header in `DialContext` to make it work.

docs/CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ When releasing a new version:
 
 - The new `optional: generic` allows using a generic type to represent optionality. See the [documentation](genqlient.yaml) for details.
 - For schemas with enum values that differ only in casing, it's now possible to disable smart-casing in genqlient.yaml; see the [documentation](genqlient.yaml) for `casing` for details.
+- genqlient now supports subscriptions
```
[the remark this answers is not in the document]
matthieu4294967296moineau: done

generate/operation.go.tmpl
```diff
@@ -36,14 +35,60 @@ func {{.Name}}(
...
+    defer close(dataChan_)
+    var gqlResp graphql.Response
+	var wsResp {{.Name}}WsResponse
+	for {
+		jsonRaw, more_ := <-respChan_
```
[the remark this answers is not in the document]
matthieu4294967296moineau: done

graphql/client.go
matthieu4294967296moineau: change made [here](https://github.com/Khan/genqlient/pull/294#discussion_r1348374734)

graphql/websocket.go
```diff
@@ -0,0 +1,155 @@
...
+	case webSocketTypeNext, webSocketTypeError:
+		respChan <- wsMsg.Payload
+	default:
+	}
+	return nil
+}
+
+// formatCloseMessage formats closeCode and text as a WebSocket close message.
+// An empty message is returned for code CloseNoStatusReceived.
+func formatCloseMessage(closeCode int, text string) []byte {
```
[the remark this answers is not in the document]
matthieu4294967296moineau: It can be used to send any custom message when closing the connection. But yes, for now, it is unused.
Tell me if you want it removed.

graphql/client.go
```diff
@@ -40,19 +42,68 @@ type Client interface {
...
+	Unsubscribe(subscriptionID string) error
+}
+
 type client struct {
 	httpClient Doer
 	endpoint   string
 	method     string
 }
 
+type webSocketClient struct {
```
benjaminjkraft, line 87: nit: probably clearer to move this + methods below to `websocket.go` (you can keep the interfaces here or not, either way)

graphql/client.go
```diff
@@ -239,3 +334,70 @@ func (c *client) createGetRequest(req *Request) (*http.Request, error) {
...
+func (w *webSocketClient) Unsubscribe(subscriptionID string) error {
+	completeMsg := webSocketSendMessage{
+		Type: webSocketTypeComplete,
+		ID:   subscriptionID,
+	}
+	err := w.sendStructAsJSON(completeMsg)
+	if err != nil {
+		return err
+	}
+	w.subscriptions.Delete(subscriptionID)
```
benjaminjkraft, line 401: Then conversely here, if we receive more messages for this sub that were already inflight when we sent the complete, we'll drop them. I think probably we want to keep the subscription ID around (maybe clean up the channel and just no-op). Then we can just discard those instead of erroring.

(Or, we could discard without error on unknown subscription ID. I don't like the idea of dropping errors like that but it's hard to know what the caller would do with them; at best they're useful to find a bug in genqlient or in the server. But probably better to just make that unlikely; we can always make the error `Is`-able if it becomes an issue with some servers.)

graphql/client.go
```diff
@@ -239,3 +334,70 @@ func (c *client) createGetRequest(req *Request) (*http.Request, error) {
...
+	subscriptionMsg := webSocketSendMessage{
+		Type:    webSocketTypeSubscribe,
+		Payload: req,
+		ID:      subscriptionID,
+	}
+	err := w.sendStructAsJSON(subscriptionMsg)
+	if err != nil {
+		return "", err
+	}
+	w.subscriptions.Create(subscriptionID, interfaceChan, forwardDataFunc)
```
benjaminjkraft, line 388: We can end up in trouble (unknown subscription ID error + dropped message) if the first message from the websocket comes in before we add the subscription to the map. I think it suffices to add the subscription to the map first (then remove it on error).

benjaminjkraft (commented):
Thanks for all the updates, and sorry for the huge delay in reviewing here! Finding the time to review bigger PRs has been tricky for me.

I think this is quite close -- most of my comments are on documentation plus a couple race conditions that are probably easy-enough to fix. (I'm not super worried about getting the implementation totally perfect there as long as the interface is workable.)

One high-level thing to look at and document in the interface is: who is supposed to close the data/error channels? Ideally we can avoid any race conditions like the unsubscribe issue, since pushing those things to callers gets pretty annoying.

HaraldNordgren:
@matthieu4294967296moineau Any updates here? It would be great to have this availible! 🤗

its-a-feature:
Any progress here? This seems super cool and I love the functionality so far, but I need subscriptions to be possible before I can fully use it :(

weidonglian:
Any progress on this? @matthieu4294967296moineau 
Thanks for your contribution and effort on this.

.golangci.yml
```diff
@@ -91,7 +92,7 @@ issues:
     # Ok to use fmt.Print in the examples, and in the CLI entrypoint.
     - linters:
         - forbidigo
-      path: ^example/|^generate/main\.go$
+      path: ^example/|^exampleSubscription/|^generate/main\.go$
```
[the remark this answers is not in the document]
matthieu4294967296moineau: reverted

graphql/client.go
matthieu4294967296moineau: done

graphql/client.go
matthieu4294967296moineau: So now I keep the subscription in the map but with an attribute `hasBeenUnsubscribed` that is set to `true` when the client unsubscribes, and that is checked when receiving a websocket message to decide whether or not forwarding the message to the channel. Does it work for you ?

graphql/client.go
matthieu4294967296moineau: done, I kept the interface here next to the `Client`

graphql/client.go
```diff
@@ -40,19 +42,68 @@ type Client interface {
...
+	//
+	// errChan is a channel on which are sent the errors of webSocket
+	// communication.
+	//
+	// err is any error that occurs when setting up the webSocket connection.
+	StartWebSocket(ctx context.Context) (errChan chan error, err error)
+
+	// CloseWebSocket must close the webSocket connection. If no connection was
+	// started, CloseWebSocket is a no-op
+	CloseWebSocket()
```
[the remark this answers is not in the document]
matthieu4294967296moineau: done

graphql/client.go
```diff
@@ -40,19 +42,68 @@ type Client interface {
...
+
+	// Subscribe must subscribe to an endpoint of the client's GraphQL API.
+	//
+	// req contains the data to be sent to the GraphQL server. Will be marshalled
+	// into JSON bytes.
+	//
+	// interfaceChan is a channel used to send the data that arrives via the
+	// webSocket connection.
```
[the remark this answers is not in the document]
matthieu4294967296moineau: done

graphql/client.go
```diff
@@ -40,19 +42,68 @@ type Client interface {
 	) error
 }
 
+type WebSocketClient interface {
+	// StartWebSocket must open a webSocket connection and subscribe to an endpoint
+	// of the client's GraphQL API.
+	//
+	// errChan is a channel on which are sent the errors of webSocket
+	// communication.
```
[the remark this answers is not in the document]
matthieu4294967296moineau: correct

graphql/client.go
```diff
@@ -239,3 +334,70 @@ func (c *client) createGetRequest(req *Request) (*http.Request, error) {
...
+
+func (w *webSocketClient) CloseWebSocket() {
+	if w.conn == nil {
+		return
+	}
+	defer w.conn.Close()
+	err := w.conn.WriteMessage(closeMessage, formatCloseMessage(closeNormalClosure, ""))
+	if err != nil {
+		w.errChan <- err
+	}
```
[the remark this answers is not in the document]
matthieu4294967296moineau: fixed 👍

generate/operation.go.tmpl
```diff
@@ -36,14 +35,50 @@ func {{.Name}}(
...
+
+func {{.Name}}ForwardData(interfaceChan interface{}, jsonRawMsg json.RawMessage) error {
+    var gqlResp graphql.Response
+	var wsResp {{.Name}}WsResponse
+    err := json.Unmarshal(jsonRawMsg, &gqlResp)
+    if err != nil {
+        return err
+    }
+    if len(gqlResp.Errors) == 0 {
+        err = json.Unmarshal(jsonRawMsg, &wsResp)
```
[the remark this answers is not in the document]
matthieu4294967296moineau: so the unmarshal into the wsResp does not work as the error in the `jsonRawMsg` as, in case of an error, the `jsonRawMsg` looks like this:
`{"errors":[{"message":"error message","path":["endpoint"],"extensions":{"code":5}}],"data":null}`

another way of doing it would be to return a `gqlerror.List` (adding a dependency) or to define our own `Error` type

docs/CHANGELOG.md
```diff
@@ -26,6 +26,8 @@ When releasing a new version:
 
 - The new `optional: generic` allows using a generic type to represent optionality. See the [documentation](genqlient.yaml) for details.
 - For schemas with enum values that differ only in casing, it's now possible to disable smart-casing in genqlient.yaml; see the [documentation](genqlient.yaml) for `casing` for details.
+- genqlient now supports subscriptions; the websocket protocol is by default `graphql-transport-ws` but can be set to another value.
```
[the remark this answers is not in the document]
matthieu4294967296moineau: done

.golangci.yml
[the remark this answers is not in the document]
matthieu4294967296moineau: done

matthieu4294967296moineau (commented):
After a long time, I finally had time to come back on this topic !
I answered the points you mentioned in your comments.

Questions:
- I had to go back to using an old version of go to generate the files. Are you planning on making it compatible with go 1.22 ?
- How do you suggest rebasing this branch ?

graphql/client.go
benjaminjkraft: That seems great!

benjaminjkraft:
Great! Thanks for the fixes and sorry for the delay! I did the merging, it looks like I can't push to your branch (not sure if that's a permission issue on my end or yours but oh well) but I'll just stick it in a new PR and merge that way.

benjaminjkraft:
Merged via #347! Thanks so much!

PR #298, opened 2023-12-20
Add option for including operationName as a query parameter

cgetzen:
This introduces the ability to include `operationName` as a query parameter for the default client.

There are a few patterns for options that are described [here](https://golang.cafe/blog/golang-functional-options-pattern.html). The existing `NewClientUsingGet` follows the "Declare a new constructor for each configuration option" pattern. As additional options get added, the "Functional Options Pattern" may be more scalable.

Use-case: This allows requests to be routed based on their operation.

I have:

### DIFF
--- a/graphql/client.go
+++ b/graphql/client.go
@@ -41,9 +41,15 @@ type Client interface {
 }
 
 type client struct {
-	httpClient Doer
-	endpoint   string
-	method     string
+	httpClient             Doer
+	endpoint               string
+	method                 string
+	withOperationNameParam bool
+}
+
+// WithOperationNameParam allows operationName to be included as a query parameter.
+func WithOperationNameParam(c *client) {
+	c.withOperationNameParam = true
 }
 
 // NewClient returns a [Client] which makes requests to the given endpoint,
@@ -58,8 +64,8 @@ type client struct {
 // example.
 //
 // [example/main.go]: https://github.com/Khan/genqlient/blob/main/example/main.go#L12-L20
-func NewClient(endpoint string, httpClient Doer) Client {
-	return newClient(endpoint, httpClient, http.MethodPost)
+func NewClient(endpoint string, httpClient Doer, options ...func(*client)) Client {
+	return newClient(endpoint, httpClient, http.MethodPost, options...)
 }
 
 // NewClientUsingGet returns a [Client] which makes GET requests to the given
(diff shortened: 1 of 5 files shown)


cgetzen:
Hi @StevenACoffman

The lint failures seem unrelated to this PR. I am also unable to trigger these errors locally with:

```
docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:v1.52.2 golangci-lint run -v
```

Do you have any guidance on what could be going wrong? Thanks!

graphql/client.go
```diff
@@ -58,8 +64,8 @@ type client struct {
 // example.
 //
 // [example/main.go]: https://github.com/Khan/genqlient/blob/main/example/main.go#L12-L20
-func NewClient(endpoint string, httpClient Doer) Client {
-	return newClient(endpoint, httpClient, http.MethodPost)
+func NewClient(endpoint string, httpClient Doer, options ...func(*client)) Client {
```
benjaminjkraft, line 67: technically a breaking change but probably one we can get away with

graphql/client.go
```diff
@@ -178,14 +192,31 @@ func (c *client) MakeRequest(ctx context.Context, req *Request, resp *Response)
...
+
 	body, err := json.Marshal(req)
 	if err != nil {
 		return nil, err
 	}
 
 	httpReq, err := http.NewRequest(
 		c.method,
-		c.endpoint,
+		parsedURL.String(),
```
benjaminjkraft, line 219: Is parse/unparse guaranteed to roundtrip a valid URL without changes? Unless we're sure, it's probably a tad safer if we change `queryUpdated` to `urlUpdated` and use the URL exactly as specified if we don't need to change it.

internal/integration/integration_test.go
benjaminjkraft, line 1: Please also add at least a quick test (either unit or integration, not sure which will be easier to wire up) that the option actually does something!

graphql/client.go
```diff
@@ -41,9 +41,15 @@ type Client interface {
...
-	endpoint   string
-	method     string
+	httpClient             Doer
+	endpoint               string
+	method                 string
+	withOperationNameParam bool
+}
+
+// WithOperationNameParam allows operationName to be included as a query parameter.
+func WithOperationNameParam(c *client) {
```
benjaminjkraft, line 51: please add a named type for the options so they show up nicely in the documentation

benjaminjkraft (commented):
Thanks for the PR! The code looks fine modulo a few comments, below.

But I do want to check: does this match any standard conventions? I don't see any specific problems here but I want to be a bit careful of going down the road of adding many options to the client. (One can always wrap the `Doer` (specifically, the `http.Transport` of the `http.Client`) instead -- indeed at Khan when we wrote genqlient we did as much to support something similar but with a slightly different convention. For `NewClientUsingGet` that's substantially harder but here it would be basically the same code.)

benjaminjkraft:
Ah, and I'm not sure what's up with the lint. I agree it looks off, but I don't have time to figure out why right now. Perhaps we should try updating the linter -- feel free to give that a go in a separate PR if you like!

benjaminjkraft:
Lint is fixed in #311 so once you merge main that part should be good!

benjaminjkraft:
Closing, as this is stale and there are other options. Feel free to make a new PR if you address the comments!

PR #316, opened 2024-02-18
Improve package-sniffing and bind correctly to types in the same package

### LINKED ISSUE #283: Binding to a type in the same package as generated code produces invalid output

benjaminjkraft:
The original motivation here is to fix #283, which is that if you try to use `bindings` to bind to a type in the same package as the generated code, we generate a self-import, which Go doesn't allow. Fixing that is easy -- the three lines in `imports.go` -- once you know the package-path of the generated code. (The test that that all fits together is in integration-tests because that was the easiest place to set up the right situation.)

Determining the package-path is not too much harder: you ask `go/packages`, which we already use for `package_bindings`. But once we're doing that, and handling errors, it's kinda silly that we ask you to specify the package your generated code will use, because we have that too, usually. So I rewrote that handling too, making `package` now rarely necessary (see for example the `example` config), and warning if it looks wrong. This is the changes in `config.go`, and is the more substantial non-test change. (I also renamed some of the testdata dirs to be valid package-names, to exercise more of that code, or in the case of `find-config`, just for consistency.)

I have:
- [x] Added tests covering my changes, if applicable (and checked that the test for the bugfix fails without the rest of the changes)
- [x] Added ~an entry~ two entries to the changelog


generate/config.go
```diff
@@ -201,6 +256,11 @@ func (c *Config) ValidateAndFillDefaults(baseDir string) error {
 					binding.Package)
 			}
 
+			if binding.Package == c.pkgPath {
+				warn(errorf(nil, "warning: package_bindings set to the same package as your generated "+
+					"code ('%v'); this will probably cause circularity issues", c.pkgPath))
```
csilvers, line 261: "probably cause circular-import issues"?

csilvers (approved):
This looks good to me!  I like figuring out the package name automatically; that makes a lot of sense to me.  Some style comments below but nothing requiring another go-around.

PR #322, opened 2024-02-19
Reorganize and improve documentation

### LINKED ISSUE #63: Figure out, and document, what semver means for us
In particular, obviously it's a breaking change if we change the signature of an exported symbol, like `graphql.Client` or `generate.Main`.  But are these breaking changes?

The first question is: do we want to treat as breaking a change to the code-generator such that the new generated code has the same API but different implementation?  If you have a test to require that `go generate ./...` doesn't make any changes, this will break that test, and that test is a reasonable thing you might want to have.  But I think realistically there's no way we can treat that as a breaking change, since probably half our commits, including plenty of bugfixes, will change some generated code you might have.

The second question is: what if the new generated code has a breaking change in its API?  This is only a breaking change once you run `go generate ./...`.  I think in many cases we will want to treat it as a breaking change; certainly if we, say, change a type-name or the query-method signature!  But what if we, make a change that's non-breaking if you use genqlient the expected way, but is breaking if you, say, use those structs for random other stuff?  We'll want to see how much of an issue that might be for us, and document whatever we intend clearly.  (This is extra true if you put your own code in the same package as the generated code, in which case there's nothing to stop you from calling internal helpers like `__unmarshalWhateverType`, which we definitely don't intend to be public API.)

Of course we don't need to figure this out perfectly until we cut a v1.0.0; until then we can just try our best and see what happens as we go.

Another issue is what code-generator versions are compatible with what runtime-library versions, but I suspect that will be a simpler issue.

### LINKED ISSUE #128: Document more clearly how to stringify large integers
  we should have a way to add json tags for bindings
 
given

```
bindings:
  bigint:
    type: int64

query GetUser($email: String!) {
  users(where: {_and: {email: {_eq: $email}, deleted_at: {_is_null: true}}}) {
    id
    email
    pwd_hash
    name
  }
}
```

generates

```
type GetUserUsers struct {
	Id       int64  `json:"-"`
	Email    string `json:"email"`
	Pwd_hash string `json:"pwd_hash"`
	Name     string `json:"name"`
}
```

graphql server sends stringified json for bigint fields which can not be parsed. 
 
If we had a way to add json tags for a binding we can add a tag `json:"id,string" to get parsed int64 back

### LINKED ISSUE #245: Documentation revamp
I think it's probably getting time for v2 of the genqlient documentation. A few thoughts:

- It might be good to have a clearer philosophy document to explain why we don't generate a single `User` Go type (or why it's not great to use your gqlgen types on the client), etc.
- The FAQ is doing way more work than it should, a lot of that should become how-to guides that people might actually be able to find.
- There are probably some newer options that should make their way into said how-to guides.
- It's probably time for more examples -- or at least to make space for contributors to add them!

Specific things to fix:
- #128

Other ideas are welcome!

benjaminjkraft:
We've been in need of a documentation revamp for a while; a giant FAQ was never the greatest structure and got worse as it grew. In this commit I reorganize the documentation. Most of it is just moving around existing text, but I added some new documentation here and there.

The changes:
- Many of the FAQ questions have moved to several new docs, on the client/runtime, handling your schema, and various operation types; the FAQ has answers to a few of the actually most frequent questions, as well as a few things that didn't fit elsewhere.
- We now have a `docs/README.md` which acts as an index, so we can just link to `/docs`.
- I lowercased the files that don't need match a GitHub convention, so it's a bit less yell-y.
- I added documentation on:
  - how we version genqlient (fixes #63)
  - newer options for optional types
  - a bit more on custom scalars and integer types (fixes #128)

Reviewers, you can see what it all looks like together [here](https://github.com/Khan/genqlient/tree/benkraft.docs-reorg/docs#readme).

Fixes #245.

I have:
- [x] Added tests covering my changes, if applicable (n/a)
- [x] Added an entry to the changelog (n/a)


StevenACoffman (approved):
nice!

Issue #327, opened 2024-04-02
Allow globs for schema/queries to be optional, if at least one file is found


stormsweeper:
**Is your feature request related to a problem? Please describe.**
When running generation, and any glob in the config does not match any files, an error is returned and generation fails. 

In our particular case, we have current SQL-based files inside dirs in `lib/models`, and we'd like to put graphql operations in files in those directories. Thus:

```yaml
operations:
- "lib/models/*/*.graqhql"
```

However, we do have some nested models (e.g. `lib/models/foo/bar`), but if we add

```yaml
- "lib/models/*/*/*.graqhql"
```
then the generation will exit unless actual files exist. 

**Describe the solution you'd like**
Ideally, this should allow some globs to fail but warn instead. This would allow us to add/remove the query files as needed, without having to also update the genqlient config.

**Describe alternatives you've considered**
We could add/remove the directory lines as we add/remove the files. Just a bit of extra work.

**Additional context**

* previous issue about nested dirs: #167 
* previous issue that introduced current behavior: #146

benjaminjkraft:
As I commented on #328 -- let's do #167 instead.

PR #328, opened 2024-04-02
Only warn if a glob does not match any files, but error if total is 0

stormsweeper:
This changes the filename expansion to only warn on empty globs, and error only if no files in total were found.

This is a strawman proposal for issue #327 

I have:
- [?] Added tests covering my changes, if applicable _No current tests exist, but I can add ones for this case and check error messages_
- [ x] Included a link to the issue fixed, if applicable
- [ -] Included documentation, for new features
- [ -] Added an entry to the changelog

### DIFF
--- a/generate/parse.go
+++ b/generate/parse.go
@@ -22,10 +22,6 @@ func getSchema(globs StringList) (*ast.Schema, error) {
 		return nil, err
 	}
 
-	if len(filenames) == 0 {
-		return nil, errorf(nil, "no schema files found in: %v", globs)
-	}
-
 	sources := make([]*ast.Source, len(filenames))
 	for i, filename := range filenames {
 		text, err := os.ReadFile(filename)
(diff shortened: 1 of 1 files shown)


generate/parse.go
```diff
@@ -123,6 +127,10 @@ func getQueries(basedir string, globs StringList) (*ast.QueryDocument, error) {
 		return nil, err
 	}
 
+	if len(filenames) == 0 {
+		return nil, errorf(nil, "no query files found in: %v", globs)
+	}
+
```
benjaminjkraft, line 133: can we do this in `expandFilenames` so the callers don't have to?

benjaminjkraft (commented):
You make a good case! Arguably a better case for `**` but this is easier. (And while reverting this after adding `**` support would be a breaking change, it's probably an okay-ish one for v0.) One small comment below.

csilvers:
> You make a good case! Arguably a better case for `**` but this is easier. (And while reverting this after adding `**` support would be a breaking change, it's probably an okay-ish one for v0.) One small comment below.

I agree `**` would be a better solution to this problem, and nice for other reasons as well.  How hard would it be to add it?

csilvers:
(There's https://github.com/bmatcuk/doublestar or https://github.com/yargevad/filepathx we could use, if we wanted.)

benjaminjkraft:
Ok, sorry, Craig convinced me: let's do double-star globs instead. (I think that will solve your issue equally well, right?)

stormsweeper:
@benjaminjkraft totally, happy to have spurred it along :)

Issue #290, opened 2023-07-28
`# @genqlient(for: "Input.nonOptionalField", omitempty: false)` not working


Anas737:
**Describe the bug**
Hey all 👋 

We are having a use-case where we want to apply `omitempty` on all fields except a mandatory graphql field.

We've tried the below code both at the top of the query and at the top of the `$input` line: 
`# @genqlient(omitempty: true)`
`# @genqlient(for: "Input.nonOptionalField", omitempty: false)`

 But we got the error:
`omitempty may only be used on optional arguments`

Shouldn't `omitempty` be considered for non optional graphql fields when set to `false`?

When using `# @genqlient(omitempty: true)` only, it works on **all** fields including non optional graphql fields and adds the ``json:"fieldname,omitempty"`` to them.

**To Reproduce**
Schema:
```graphql
input Input {
    requiredField: Type!
    optionalField: Type
}
```

Query:
```graphql
# @genqlient(omitempty: true)
# @genqlient(for: "Input.requiredField", omitempty: false)
mutation Name(
  $input: Input!
) {
  name(input: $input) {
     ...
  }
}
```

Happens at code generation running:
`go run -mod=readonly github.com/Khan/genqlient@v0.5.0`

**Expected behavior**
When using the below code at the top of the query:
`# @genqlient(omitempty: true)`
`# @genqlient(for: "Input.nonOptionalField", omitempty: false)`

I would expect `omitempty` to be added to all fields except the `Input.nonOptionalField` one.
I tested and it worked when using only optional fields.

**genqlient version**
Tested on both 0.5.0 and 0.6.0

benjaminjkraft:
Thanks for raising this. I agree this is a bug. We can probably just change the validation ([here](https://github.com/Khan/genqlient/blob/05817543f5ab8d34036c70b1402d234db880f8a7/generate/genqlient_directive.go#L217) and a few 
similar places in the same file) to only reject `omitempty: true`, not `omitempty: false` on non-nullable fields. Or, more correct (although probably more work) would be to just change the behavior so a global `omitempty` only applies to non-nullable fields; I don't see how that behavior could be useful.

Anas737:
Thank you for looking into it

mwajeeh:
I want it to add omitempty to everything but `# @genqlient(omitempty: true)` is throwing `omitempty may only be used on optional arguments`. Here is my mutaion:

```
# @genqlient(omitempty: true)
[...]

Issue #228, opened 2022-10-13
Default graphql values are ignored when send request


kateile:
**Describe the bug**
Hi geeks, Thanks for this amazing lib. I have the following graphql input with default fields 

```gql
input FormInput {
    hint: String
    inputType: MwappFormInputType = TEXT
    key: String = "form"
    label: String = "Fill form"
    maxLength: Int = 100
    multiSubmission: Boolean = false
    multiline: Boolean = false
    obscure: Boolean = false
}

input SendMessageInput {
    form: FormInput
}

type Mutation {
    sendMessage(connectionId: ID!, input: SendMessageInput!): String!
}
```

Which I use in golang like 
```go
mw.SendMessage(ConnectionId, mwapp.SendMessageInput{
		Message: "Hi, Please tell us your name. \n you can write it below",
		Form: &mwapp.FormInput{
			Key: FormKeyName.String(),
		},
	})
```

As you can see above I just use one field `Key` and ignore the rest. So I expect the default graphql fields specified in schema to be used when send request. But they are sent like null too.

**To Reproduce**
Create mutation whose input have fields with default fields

**Expected behavior**
Default fields to be respected when. I believe having pointer with `omitempty` in the generated code will be the solution. But I don't know how to archive this

**genqlient version**
v0.5.0

**Additional context**
My genqclient.yaml is 

```yaml
schema: schema.graphql
package: mwapp
operations:
- operations/*.graphql
generated: generated.go
use_struct_references : true
optional: pointer
bindings:
  Timestamp:
    type: time.Time
```

Any help on fixing this will be appreciated.

benjaminjkraft:
Huh, I didn't realize input-object fields could have defaults! (For my future self: it's at [InputObjectTypeDefinition](https://spec.graphql.org/October2021/#InputObjectTypeDefinition) which refers to [InputValueDefinition](https://spec.graphql.org/October2021/#InputValueDefinition).)

Does it work to use genqlient's [`omitempty` option](https://github.com/Khan/genqlient/blob/main/docs/FAQ.md#-nullable-fields)? For example you could add at the top of the file `# @genqlient(for: "FormInput.inputType", omitempty: true)`, and so on for each field. (Obviously that's a bit verbose; we could talk about if there's a better way to do it, perhaps by extending #198 to allow an `omitempty` option.)

Fazt01:
Hi, yes the `omitempty` works -> the field is omitted in request and the server takes the default as value for that input field.

[...]

PR #338, opened 2024-05-29
omitempty validation on input fields with/without defaults

Fazt01:
Fixes https://github.com/Khan/genqlient/issues/290 and https://github.com/Khan/genqlient/issues/228#issuecomment-2023396706 (for latter, only the comment of mine, not the whole issue)

This is only changing where the `omitempty` is allowed and forbidden - it does not change where is the `omitempty` actually generated or not in generated code.

Separately each line of changelog:
- allow `omitempty` on non-nullable input field, if the field has a default (pretty much https://github.com/Khan/genqlient/issues/228#issuecomment-2023396706)
  - added a `&& field.DefaultValue == nil` to the `"omitempty may only be used on optional arguments"` error
- allow `omitempty: false` on an input field, even when it is non-nullable
  - `fieldDir.Omitempty != nil` changed to `fieldOptions.GetOmitempty()`
- forbid `omitempty: false` (including implicit behaviour) when using pointer on non-null, no-default input field
  - as setting a correct combination of directives (and potentially some options, which changes implicit pointer and/or omitempty), and this library promises "Compile-time validation of GraphQL queries: never ship an invalid GraphQL query again!", I found it fitting to guard against the most simple case, that can be enforced in Go type system.
  - this, however, is a breaking change, so not sure if I should include it here. No previously present test failed after such change, but for example `generate/testdata/errors/DefaultInputsNoOmitPointerForDirective.graphql` would previously generate following (below - which has a possibility to send invalid graphql input), but now the generation fails.
 ```
        type InputWithDefaults struct {
        	Field         *string `json:"field"`
        	NullableField string  `json:"nullableField"`
        }
```
- - alternative would be to force omitempty tag in such cases (even if there is no omitempty directive/option) - so that generation would not fail. But I'm not sure if I can afford to do that. That would probably still be breaking change (different generated code for same query), but a bit better. Maybe just setting omitempty flag instead of returning error would be sufficient.
  
In general, I have also moved the omitempty check from directives to the time of creating Go types/tags of field. This seems more consistent, as not all possibilities were caught before (i.e. general `@genqlient(omitemtpy: true)` vs `@genqlient(for: ..., omitempty: true)`). When creating Go type/tags, all the options/directives are already merged, so the final result is being checked. There is minor difference in error message (instead of reference to directive, the error refers to the whole operation, but also includes type.field name)

I have:


benjaminjkraft (commented):
Man, this code is getting confusing. Anyway your changes seem convincing, and the tests help a lot.

I think the breaking change is a sensible one -- either the query is broken (in which case it's good to reject it) or you effectively always set the value (in which case the fix is simple and a good idea). And it must be rare-ish if none of our tests cover it. If we hear from users we can always adjust!

This should be good to merge once I get confirmation from the CLA person (tomorrow, probably).

PR #354, opened 2024-09-27
Fix issue where subscription server data channel closing gives error

HaraldNordgren:
Handles errors discussed here: https://github.com/Khan/genqlient/pull/347#issuecomment-2376667291

I have:

### DIFF
--- a/graphql/websocket.go
+++ b/graphql/websocket.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"reflect"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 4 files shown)


benjaminjkraft (commented):
Seems reasonable to me. It's kinda funny we end up with reflect for this, but it seems fine, and better than adding a bunch of new plumbing to generate a single `close` call.

I'll wait a couple days in case @matthieu4294967296moineau has anything to add.

HaraldNordgren:
Yeah, I don't love the reflect part, although it does solves the problem. The reason I still went with it was because I found the same thing in subscriptionsMap.Unsubscribe().

If we really don't want to reflect then a type cast could likely also solve the problem 🤗

benjaminjkraft:
Ah, I guess it's not the first then!

A cast would be better, but won't work because we don't know the concrete type, right?

HaraldNordgren:
Ah, right!

matthieu4294967296moineau:
> Ah, I guess it's not the first then!
> 
> A cast would be better, but won't work because we don't know the concrete type, right?

yes this is why I used it in a first place to close the channel when unsubscribing

The fix looks good to me, I could test it and it worked fine 👍

HaraldNordgren:
@benjaminjkraft @matthieu4294967296moineau  Great! Thanks for the feedback!

Let's merge? 🚀

PR #360, opened 2024-11-14
Subscription connection parameters support

HaraldNordgren:
Adds support for subscription (websocket) connection parameters support.

We are having an issue with authorization when using subscriptions in genqlient. Our mobile app team is using Apollo for GraphQL requests, and Apollo seems to sends its auth in a way that is different from how genqlient does it:

https://www.apollographql.com/docs/react/data/subscriptions#5-authenticate-over-websocket-optional

This is causing a problems, because we use genqlient for testing the backend. It seems that the way Apollo is implemented, authentication (and other headers) are not actually sent as HTTP headers, but as connection parameters instead.

I have:

### DIFF
--- a/graphql/client.go
+++ b/graphql/client.go
@@ -132,6 +132,16 @@ func NewClientUsingGet(endpoint string, httpClient Doer) Client {
 // The client does not support queries nor mutations, and will return an error
 // if passed a request that attempts one.
 func NewClientUsingWebSocket(endpoint string, wsDialer Dialer, headers http.Header) WebSocketClient {
+	return NewClientUsingWebSocketWithConnectionParams(endpoint, wsDialer, headers, nil)
+}
+
+// NewClientUsingWebSocketWithConnectionParams returns a [WebSocketClient] which makes subscription requests
+// to the given endpoint using webSocket. It allows to pass additional connection parameters
+// to the server during the initial connection handshake.
+//
+// connectionParams is a map of connection parameters to be sent to the server
+// during the initial connection handshake.
+func NewClientUsingWebSocketWithConnectionParams(endpoint string, wsDialer Dialer, headers http.Header, connParams map[string]interface{}) WebSocketClient {
 	if headers == nil {
 		headers = http.Header{}
 	}
@@ -141,6 +151,7 @@ func NewClientUsingWebSocket(endpoint string, wsDialer Dialer, headers http.Head
 	return &webSocketClient{
 		Dialer:        wsDialer,
 		Header:        headers,
+		connParams:    connParams,
 		errChan:       make(chan error),
 		endpoint:      endpoint,
 		subscriptions: subscriptionMap{map_: make(map[string]subscription)},
--- a/graphql/client.go
+++ b/graphql/client.go
@@ -126,35 +126,40 @@ func NewClientUsingGet(endpoint string, httpClient Doer) Client {
 	return newClient(endpoint, httpClient, http.MethodGet)
 }
 
+type WebSocketOption func(*webSocketClient)
+
 // NewClientUsingWebSocket returns a [WebSocketClient] which makes subscription requests
 // to the given endpoint using webSocket.
 //
 // The client does not support queries nor mutations, and will return an error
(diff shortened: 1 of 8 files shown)


graphql/client.go
```diff
@@ -132,6 +132,16 @@ func NewClientUsingGet(endpoint string, httpClient Doer) Client {
...
+	return NewClientUsingWebSocketWithConnectionParams(endpoint, wsDialer, headers, nil)
+}
+
+// NewClientUsingWebSocketWithConnectionParams returns a [WebSocketClient] which makes subscription requests
+// to the given endpoint using webSocket. It allows to pass additional connection parameters
+// to the server during the initial connection handshake.
+//
+// connectionParams is a map of connection parameters to be sent to the server
+// during the initial connection handshake.
+func NewClientUsingWebSocketWithConnectionParams(endpoint string, wsDialer Dialer, headers http.Header, connParams map[string]any) WebSocketClient {
```
HaraldNordgren, line 144: Since there was never a a release that includes `NewClientUsingWebSocket`, we should be free to update the signature without breaking anything and possible have only one of these:

```go
func NewClientUsingWebSocket(endpoint string, wsDialer Dialer, headers http.Header, connParams map[string]any) WebSocketClient
```

HaraldNordgren:
A big gotcha regarding HTTP headers and connection parameters, that I realized recently, is that websocket requests in 99desgins gqlgen follow a different path, going through here:

```go
	gqlgenServer.AddTransport(transport.Websocket{
		InitFunc: func(ctx context.Context, initPayload transport.InitPayload) (context.Context, *transport.InitPayload, error) {
			// initPayload contains connection parameters
		},
	})
```

HaraldNordgren:
Ping @benjaminjkraft and @matthieu4294967296moineau, do you think it would be possible to get a quick look here? 🤗

HaraldNordgren:
Ping @benjaminjkraft 🤗

jbaldwinroberts:
this would be useful for us too - thanks

HaraldNordgren:
Hi @benjaminjkraft, I know you are likely very busy. But it would be much appreciated to get some review here 🤗

benjaminjkraft (commented):
Sorry for the delay, it's been a busy month. This seems ok, but I wonder if now is the right time to add a more extensible options interface, either an options struct, or functional-style options, either of which could incorporate the optional headers as well. That way we don't end up having to add a ton of different functions over time.

HaraldNordgren:
@benjaminjkraft Hope everything is under control and not too stressful 🤗

Thanks for getting a review here, much appreciated!

I think it's a great point regarding options. I have pushed a new version with this added. Let me know what you think!

HaraldNordgren:
I couldn't figure out a way to generate the 99designs/gqlgen code for countAuthorized. Running 'go generate ./...' like I would do for my regular repositories fails.

So I created countAuthorized infrastructure myself. I hope  didn't introduce any issue. All the tests are passing at least 😃

benjaminjkraft (commented):
Huh, `go generate ./...` should work! It works for me. Adding `-v -x` may make any issues clearer, you can see what packages and commands it's finding. Anyway, you got it right except for one line of whitespace :-) .

On the actual changes, this looks great. I'm not sure if we should change headers to be an option as well, since it's optional. I think I would incline towards yes. As long as it's a 1:1 replacement we can probably just change the existing API since as you say there hasn't been a release, then we don't have to have two. But if you see a reason not to I'm also ok to merge this as-is.

HaraldNordgren:
@benjaminjkraft I think it does make sense with the headers.

But it feels a bit tricky to get it right with the important "Sec-WebSocket-Protocol" key, which should always be present in the output. Maybe better to do as a follow-up?

benjaminjkraft:
Oh, I see what you mean. Sure, I'll merge this and if we sort it out before we cut a release great, and if not oh well.

HaraldNordgren:
Thanks a lot @benjaminjkraft 🚀

benjaminjkraft:
Thank you!!

PR #362, opened 2024-11-30
Update Go (and golangci-lint, and gqlgen) versions, and test codegen

benjaminjkraft:
Originally my goal here was to add a check that generated code is up to date. (In #360 I was reminded that we have that for first-party code, but not gqlgen.)

Then I figured I'd update the Go version to support 1.23, which required updating gqlgen and golangci-lint, both of which now have minimum Go versions of 1.22. So now that's ours too. Those are the only upstream supported versions anyway, so hopefully people are updated.

I suspect the CI version-setting stuff can get simplified a bit with all the Go toolchain version stuff, but for now I didn't bother.

I have:


csilvers (approved):
Seems good to me!

PR #352, opened 2024-09-20
Standardize error return from HTTP requests in genqlient

wwzeng1:
# Purpose
This pull request aims to standardize the error handling for HTTP requests in the genqlient library. The current implementation returns a generic error message that makes it difficult to handle specific error scenarios, such as retrying on rate limit errors.

Fixes #333

I have:
- [x] Written a clear PR title and description (above)
- [x] Signed the [Khan Academy CLA](https://www.khanacademy.org/r/cla)
- [x] Added tests covering my changes, if applicable
- [x] Included a link to the issue fixed, if applicable
- [ ] Included documentation, for new features
- [ ] Added an entry to the changelog

PR #363, opened 2024-11-30
Standardize error return from HTTP requests

### LINKED ISSUE #333: Standardise the error return from http request
Hey there. I want this because I need to do a error handling of the gql request. Situations like, if the request is returning 429 error, the program should do a retry, etc. However the current code is returning` fmt.Errorf("returned error %v: %s", httpResp.Status, respBody)`. It is not helping the situation. It makes error handling complicated. Can we return the standard error so that can help the user better handling the errors?

We can do a simple enum. Or we can create a new error object includes the http status code + error info (that includes the resp body). 

It will be really nice if we can have such feature. Otherwise I will just fork the code and create my derivatives. Thanks.

benjaminjkraft:
This makes the HTTP request errors `As`-able, so you can retry on rate limit errors, or whatever.

Fixes #333, replaces #352 (I just added docs).

I have:

### DIFF
--- a/graphql/errors.go
+++ b/graphql/errors.go
@@ -4,8 +4,8 @@ import "fmt"
 
 // HTTPError represents an HTTP error with status code and response body.
 type HTTPError struct {
-	StatusCode int
 	Body       string
+	StatusCode int
 }
 
 // Error implements the error interface for HTTPError.
--- a/graphql/errors.go
+++ b/graphql/errors.go
@@ -1,0 +1,14 @@
+package graphql
+
+import "fmt"
+
+// HTTPError represents an HTTP error with status code and response body.
+type HTTPError struct {
+	StatusCode int
+	Body       string
+}
+
+// Error implements the error interface for HTTPError.
+func (e *HTTPError) Error() string {
+	return fmt.Sprintf("returned error %v: %s", e.StatusCode, e.Body)
+}
--- a/docs/CHANGELOG.md
+++ b/docs/CHANGELOG.md
@@ -31,6 +31,7 @@ When releasing a new version:
   See the [documentation](FAQ.md) for how to `subscribe to an API 'subscription' endpoint`.
 - genqlient now supports double-star globs for schema and query files; see [`genqlient.yaml` docs](genqlient.yaml) for more.
 - genqlient now generates slices containing all enum values for each enum type.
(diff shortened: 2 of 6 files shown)


graphql/errors.go
```diff
@@ -0,0 +1,14 @@
...
+
+// HTTPError represents an HTTP error with status code and response body.
+type HTTPError struct {
+	Body       string
+	StatusCode int
+}
+
+// Error implements the error interface for HTTPError.
+func (e *HTTPError) Error() string {
+	return fmt.Sprintf("returned error %v: %s", e.StatusCode, e.Body)
```
csilvers, line 13: This error-text can be hard to parse when the body is the empty string (or just whitespace), so you may want to do `%#v` instead of `%s`, or something like `>>>%s<<<`.  Though maybe it's not going to matter much in practice.

csilvers (commented):
I'm late to the party, but this looks good to me!  I appreciate being able to introspect on errors.

graphql/errors.go
benjaminjkraft: I think it needs to be `%s` to convert the bytes to string, but I can at least add quotes!

HaraldNordgren:
Hi @benjaminjkraft!

This is really nice, but I wonder if we can go one step longer. This is how my tests were changed to fit with these latest changes:

```diff
-                       expected: errors.New("returned error 422 Unprocessable Entity: {\"errors\":[{\"message\":\" is not a valid MealType\",\"path\":[\"variable\",\"request\",\"mealType\"],\"extensions\":{\"code\":\"GRAPHQL_VALIDATION_FAILED\"}}],\"data\":null}"),
+                       expected: &graphql.HTTPError{
+                               Body:       "{\"errors\":[{\"message\":\" is not a valid MealType\",\"path\":[\"variable\",\"request\",\"mealType\"],\"extensions\":{\"code\":\"GRAPHQL_VALIDATION_FAILED\"}}],\"data\":null}",
+                               StatusCode: 422,
+                       },
```

Could we further and expose the underlying `gqlerror.Error` type instead?

```go
type Error struct {
	Err        error                  `json:"-"`
	Message    string                 `json:"message"`
	Path       ast.Path               `json:"path,omitempty"`
	Locations  []Location             `json:"locations,omitempty"`
	Extensions map[string]interface{} `json:"extensions,omitempty"`
	Rule       string                 `json:"-"`
}
```

Issue #371, opened 2025-02-25
schema did not match any files under win cmd


RobinKamps:
**Describe the bug**
The code generation does not work under Windows / cmd or powershell.
It worked a few Versions before - now i get always the error (on cmd or powershell):

```
go run github.com/Khan/genqlient
C:\dev\monorepo\golang\apps\license\license\license.graphqls did not match any files
```

It does not matter if the file is in the same folder or another folder - e.g.
`schema: ../license/license/license.graphqls`
or
`schema: ./license.graphqls`
or
`schema: ./license.graphql`

However it does work in a Lunix Subsystem under Windows.

**To Reproduce**
try to generate under windows

**Expected behavior**

Code generateion should find the schema under Win, too

**genqlient version**
latest

benjaminjkraft:
Fixed in #373, I think!

0xHexE:
Doesn't work I tried with

	github.com/Khan/genqlient v0.8.0 // indirect

kunibald413:
also get this issue

kunibald413:
had too add the two line form the pull request in here `go\pkg\mod\github.com\!khan\genqlient@v0.8.0\generate\parse.go`
manually
then it worked

benjaminjkraft:
Ah, I think #373 is not in any release yet. If you use latest main, does that work? I can try and cut a release, but it won't be today.

kunibald413:
i can get by like this, thanks. dunno how much hassle it is, but i guess would be good to put into a release. 
can confirm that these two lines fix it for me on windows, thanks again!

benjaminjkraft:
Ok, this is now in `v0.8.1`!

PR #373, opened 2025-02-28
Matching in genqlient by Normalizing Paths

imbytecat:
This pull request addresses an issue with the `genqlient` library, which uses the `doublestar` package for path matching. The current implementation does not correctly match paths on Windows systems due to differences in path separators. I've implemented a simple fix by converting Windows-style backslashes (`\`) to forward slashes (`/`) before performing the match, allowing the path matching to work correctly across all platforms.

#371
bmatcuk/doublestar#61


benjaminjkraft (approved):
I don't really have a good way to test this but it seems plausible!

PR #376, opened 2025-04-13
Add field aliasing

### LINKED ISSUE #367: Customize field names without alias
By default ugly names will be generated if the field name contains an underscore. This can be worked around using aliases as [documented](https://github.com/Khan/genqlient/blob/main/docs/operations.md#field-names), but this has limitations. If the server operator has configured [max aliases](https://escape.tech/graphql-armor/docs/plugins/max-aliases/), queries will fail when they exceed this.

The ability to customize the field names using a `@genqlient` directive as an alternative to aliases.

clinta:
- Added @genqlient(alias) directive to customize field names without GraphQL aliases (fixes #367)

I have:


csilvers:
Hi @clinta , thank you for your contribution!  Can you separate the last item ("Fixed path handling in file globbing to properly normalize paths") into its own PR?  I suspect that will be easier/faster to get in than the others.

Ideally, you'd have three PRs, so we could consider the camel-case and the alias changes separately.  I know those might be more intertwined, but hopefully it's not that much work.

clinta:
That's my mistake on the changelog. I was reviewing and summarizing differences between main, but I was looking at my main which hadn't pulled recent changes, after I already rebased this branch. So when I was reviewing changes, that looked like a change that needed to be included in the changelog for this PR. I fixed that.

I should be able to split up the alias and snake_case option into separate PRs.

clinta:
This is now a focused PR just on adding the alias directive. I'll open another PR based on top of this with the snake_case conversion feature.

benjaminjkraft:
Thanks for splitting this out, it's very clean now! (I'm still sad we need it, but oh well, that's life being a client.)

PR #377, opened 2025-04-13
Auto snake case

clinta:
This builds on top of #376 by adding a configuration option to automatically convert all snake_case type and field names to CamelCase.

This PR is a draft until #376 is merged.

I have:


docs/genqlient.yaml
```diff
@@ -243,6 +243,22 @@ package_bindings:
...
+# This is useful when working with GraphQL servers that use snake_case naming
+# conventions, while maintaining idiomatic Go struct field and type names.
+#
+# For example, with this option enabled:
+# - A GraphQL field named "user_id" will be named "UserId" in the generated Go struct
+# - A GraphQL type named "user_profile" will be named "UserProfile" in Go
+# - All original names are preserved in the GraphQL request
+#
+# Defaults to false.
+auto_camel_case: false
```
benjaminjkraft, line 260: let's put this under `casing` below? I think the thing that makes sense is to add a `casing.default` with the same values that controls what we do "everywhere", and then support `auto_camel_case` as a value for each. (I'm guessing that's not too much work?) So it would look like
```
casing:
  default: auto_camel_case
```
for what you have, but then someone could also do
```
casing:
  default: auto_camel_case
  all_enums: raw
```
or whatever and have them work together correctly

generate/convert.go
```diff
@@ -171,7 +171,11 @@ func (g *generator) convertArguments(
 			return nil, err
 		}
 
-		goName := upperFirst(arg.Variable)
+		goName := arg.Variable
+		if g.Config.AutoCamelCase {
+			goName = snakeToCamel(goName)
+		}
+		goName = upperFirst(goName)
```
benjaminjkraft, line 178: may be worth breaking this into a method now? especially since that will make it easy to add support for `default: raw` as I describe in the config

generate/names_test.go
```diff
@@ -60,9 +60,58 @@ func TestTypeNames(t *testing.T) {
...
+		false,
+	}, {
+		// With auto_camel_case
+		"ServiceIPsIpAddressListSnakeCaseType",
+		[]*ast.Field{fakeField("Query", "ip_address_list")},
+		"snake_case_type",
+		true,
+	}, {
+		// With nested snake_case fields
+		"ServiceIPsObjectSnake_case_fieldSnake_case_type",
```
benjaminjkraft, line 95: these are very convincing!

benjaminjkraft (commented):
Cool, this looks pretty good. The one thing I'd like to have is for it to integrate with existing config better -- the goal being to avoid the situation where we have a bunch of overlapping but distinct config options in the future. I think that won't be too much work; it's mostly just plumbing things differently I suspect. But if it turns out to be too much trouble we could do a subset of that for now.

clinta:
Take a look at this, I've tired to refactor to a more unified approach of applying casing algorithms. If this looks good, I can squash commits before you merge.

benjaminjkraft:
This is great, thank you! I already squash when merging normally so it should be fine.

PR #384, opened 2025-06-11
[Subscription] check already closed channels 

cheina97:
This PR fixes an issue I encountered using subscriptions with genqlient.

I've noticed that the Unsubscribe function closes the subscription channel, but never checks if that channel has already been closed. 

So a snippet like that causes a panic when Close is called:

``` go
err = gqlsubcl.Unsubscribe(subscriptionID)
if err != nil {
	return fmt.Errorf("failed to unsubscribe from GraphQL subscription: %w", err)
}

if err := gqlsubcl.Close(); err != nil {
	return fmt.Errorf("failed to close GraphQL subscription client: %w", err)
}
```

I have:

### DIFF
--- a/graphql/subscription.go
+++ b/graphql/subscription.go
@@ -44,9 +44,13 @@ func (s *subscriptionMap) Unsubscribe(subscriptionID string) error {
 	if !success {
 		return fmt.Errorf("tried to unsubscribe from unknown subscription with ID '%s'", subscriptionID)
 	}
+	hasBeenUnsubscribed := unsub.hasBeenUnsubscribed
 	unsub.hasBeenUnsubscribed = true
 	s.map_[subscriptionID] = unsub
-	reflect.ValueOf(s.map_[subscriptionID].interfaceChan).Close()
+
+	if !hasBeenUnsubscribed {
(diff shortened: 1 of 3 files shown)


graphql/subscription.go
```diff
@@ -44,9 +44,13 @@ func (s *subscriptionMap) Unsubscribe(subscriptionID string) error {
 	if !success {
 		return fmt.Errorf("tried to unsubscribe from unknown subscription with ID '%s'", subscriptionID)
 	}
+	hasBeenUnsubscribed := unsub.hasBeenUnsubscribed
 	unsub.hasBeenUnsubscribed = true
 	s.map_[subscriptionID] = unsub
-	reflect.ValueOf(s.map_[subscriptionID].interfaceChan).Close()
+
+	if hasBeenUnsubscribed {
```
benjaminjkraft, line 51: Isn't this backwards?
```suggestion
	if !hasBeenUnsubscribed {
```

Maybe a test would help confirm, although I'm not sure how hard that is to set up, maybe it's not worthwhile.

benjaminjkraft (approved):
Thanks!

PR #393, opened 2025-08-02
Fix test flake revealed by #308

### LINKED ISSUE #281: Split `TestGenerateWithConfig` cases
When we run several queries against the same snapshot, we dump them all into one genqlient run. Would probably be clearer to do one run per query; this confused me for a bit while reviewing #280.

benjaminjkraft:
Two issues here:
- In the with-config tests, we were generating all the different input files into one output. This makes sense in user code, but in our tests, we reuse the same types all over the place, so not only is it confusing to read all the tests interleaved, it can cause nondeterminism anytime our conflict-detection is incomplete (#123). As it happens, #308 introduced such a case, so tests started flaking. Fixes #281.
- This turned the case introduced by #308 from a flake into a consistent failure. The issue is actually unrelated to the changes in the PR (it's intentional validation added in #338, and specifically corresponds to `errors/OmitemptyDirective.graphql` from that PR), so I changed the config for the test case to one that avoids the issue while still testing what we want.

I have:


csilvers (commented):
I understand how this PR fixes the first issue you mention in the commit message, but I'm not sure I understand the second.  I'd like to hold off approving until I do.

I'm guessing that the change to add PokemonInput is what fixes the second issue?  How does that fix anything?

benjaminjkraft:
Yes, sorry, the second fix is to add the PokemonInput type to `bindings`. That means instead of autogenerating a Go type for that GraphQL type, we use the existing one. Generating the type hit one of the cases we now intentionally validate against (#338 -- where a global `omitempty` applies to a field that doesn't make sense for); we could fix it in other ways (e.g. an explicitly `omitempty: false` on that field) but this was more consistent with how the other tests treat this type, and I believe we already have tests for those other cases anyway.

csilvers (approved):
ok, sounds good!

PR #407, opened 2025-10-28
websocket: prevent deadlocks

HaraldNordgren:
Follow-up to https://github.com/Khan/genqlient/pull/402#issuecomment-3454733476.

### DIFF
--- a/graphql/websocket.go
+++ b/graphql/websocket.go
@@ -131,6 +131,26 @@ func (w *webSocketClient) listenWebSocket() {
 	}
 }
 
+func (w *webSocketClient) getSubscriptionOrHandleComplete(subscriptionID string, subscriptionType string) (*subscription, error) {
+	w.subscriptions.Lock()
+	defer w.subscriptions.Unlock()
+	sub, success := w.subscriptions.map_[subscriptionID]
+	if !success {
+		return nil, fmt.Errorf("received message for unknown subscription ID '%s'", subscriptionID)
+	}
+	if sub.hasBeenUnsubscribed {
+		return nil, nil
+	}
+	if subscriptionType == webSocketTypeComplete {
+		sub.hasBeenUnsubscribed = true
+		w.subscriptions.map_[subscriptionID] = sub
+		reflect.ValueOf(sub.interfaceChan).Close()
+		return nil, nil
+	}
+
+	return &sub, nil
+}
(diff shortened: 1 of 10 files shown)


graphql/websocket.go
```diff
@@ -139,20 +138,9 @@ func (w *webSocketClient) forwardWebSocketData(message []byte) error {
...
-	}
-	if sub.hasBeenUnsubscribed {
-		return nil
-	}
-	if wsMsg.Type == webSocketTypeComplete {
-		sub.hasBeenUnsubscribed = true
-		w.subscriptions.map_[wsMsg.ID] = sub
-		reflect.ValueOf(sub.interfaceChan).Close()
-		return nil
+	sub, err := w.subscriptions.GetOrHandleCompletion(wsMsg.ID, wsMsg.Type)
```
HaraldNordgren: Move this into a function so that the lock is not kept when doing `sub.forwardDataFunc`. Hopefully this will prevent the deadlock.

HaraldNordgren: Scratch that, this made it worse!

HaraldNordgren:
Hi @benjaminjkraft, I'm at a bit of a loss here.

The deadlock in the tests seems going be stemming from the forwardData will still holding a lock.

However, deleting it exposes a situation where we can end up writing on a closed channel. That gives a panic which could be caught and ignored. But that seems like a very bad way to handle problems.

There seems to be an inherent race condition between checking isClosing and actually using channel. It's seems tricky to actually make this atomic.

Do you have any ideas?

benjaminjkraft (commented):
Hmm, this gets kinda complicated. I would need to think some more to see if I think it's right (for example I sure don't recall the rules for atomics). It feels like it's just sweeping problems under the rug.

I was too lazy to think so I gave this to Claude, who refactored to have just one lock (in the client); this passes the tests at least and seems less likely to have a hidden deadlock, although I'm not sure if I fully understand what invariants we need to ensure. Does having two actually let us be meaningfully more granular? (Or, can we ensure we always take the two locks out in some order? And we should probably regardless ensure we never hold a lock during network operations or calls to user-controlled code.)

HaraldNordgren:
@benjaminjkraft Makes sense, I added some comment on the PR. Thanks for the help! ❤️

PR #410, opened 2025-10-30
modernize loops

HaraldNordgren:


csilvers:
Looks like this feature was introduced in go1.23, which is the oldest version we still support.  So I'm game for this!  But I'll let @benjaminjkraft weigh in in case there's something I'm missing.

PR #412, opened 2025-11-19
Avoid deadlocks by consolidating locks

benjaminjkraft:
Alternate approach to fixing #402 -- I haven't fully checked if I think this is correct yet either but figured it was at least useful to put up as a sketch for discussion in #407.

TODO: changelog, do we want the integration test changes from #407 as well

I have:
- [ ] Written a clear PR title and description (above)

### DIFF
--- a/graphql/subscription.go
+++ b/graphql/subscription.go
@@ -1,63 +1,80 @@
 package graphql
 
 import (
-	"fmt"
 	"reflect"
-	"sync"
 )
 
-// map of subscription ID to subscription
+// subscriptionMap is a map of subscription ID to subscription.
+// It is NOT thread-safe and must be protected by the caller's lock.
 type subscriptionMap struct {
 	map_ map[string]subscription
-	sync.RWMutex
 }
 
 type subscription struct {
-	interfaceChan       interface{}
-	forwardDataFunc     ForwardDataFunction
-	id                  string
-	hasBeenUnsubscribed bool
+	interfaceChan   interface{}
+	forwardDataFunc ForwardDataFunction
+	id              string
+	closed          bool // true if the channel has been closed
 }
 
-func (s *subscriptionMap) Create(subscriptionID string, interfaceChan interface{}, forwardDataFunc ForwardDataFunction) {
-	s.Lock()
-	defer s.Unlock()
+// create adds a new subscription to the map.
+// The caller must hold the webSocketClient lock.
+func (s *subscriptionMap) create(subscriptionID string, interfaceChan interface{}, forwardDataFunc ForwardDataFunction) {
 	s.map_[subscriptionID] = subscription{
-		id:                  subscriptionID,
-		interfaceChan:       interfaceChan,
-		forwardDataFunc:     forwardDataFunc,
-		hasBeenUnsubscribed: false,
+		id:              subscriptionID,
+		interfaceChan:   interfaceChan,
+		forwardDataFunc: forwardDataFunc,
+		closed:          false,
 	}
 }
 
-func (s *subscriptionMap) Unsubscribe(subscriptionID string) error {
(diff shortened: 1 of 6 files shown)


graphql/websocket.go
```diff
@@ -108,13 +107,18 @@ func (w *webSocketClient) handleErr(err error) {
...
 		w.errChan <- err
 	}
 }
 
 func (w *webSocketClient) listenWebSocket() {
 	for {
-		if w.isClosing {
+		w.Lock()
+		isClosing := w.isClosing
+		w.Unlock()
```
HaraldNordgren, line 120: Isn't it better to always follow this pattern?

```go
w.Lock()
defer w.Unlock()

isClosing := w.isClosing
```

HaraldNordgren (approved):
Sounds reasonable. However, the actual problem has been hard for me to reproduce consistently.

I would suggest re-running all the CI 10 times, and if it passes all of those then I would consider it solved, I was never able to get 10 passes in a row (for Go 1.23, 1.24, 1.25 so 30 test runs) to pass.

Easy way to re-trigger all the CI: ```git commit --amend --no-edit --date="now" && git push -f```

graphql/websocket.go
csilvers: I don't think that will work here since the lock is happening inside a loop.  The `defer` won't run until the entire function exits, which means that the second time through the loop the Lock() would block since the lock is already being held from the first time.

You could make a separate function that just assigns `w.isClosing`, and use `defer` in that, but that's possibly overkill for this.

graphql/subscription.go
```diff
@@ -1,63 +1,80 @@
...
+// closeChannel closes a subscription's channel if it hasn't been closed yet.
+// The caller must hold the webSocketClient lock.
+// Returns true if the channel was closed, false if it was already closed.
+func (s *subscriptionMap) closeChannel(subscriptionID string) bool {
+	sub := s.get(subscriptionID)
+	if sub == nil || sub.closed {
+		return false
+	}
+
+	// Mark as closed before actually closing to prevent double-close
```
csilvers, line 73: Is this really needed given you're in a lock for this whole function?  In any case, I don't think it hurts.

graphql/websocket.go
```diff
@@ -139,22 +143,31 @@ func (w *webSocketClient) forwardWebSocketData(message []byte) error {
...
+		w.subscriptions.closeChannel(wsMsg.ID)
+		w.subscriptions.delete(wsMsg.ID)
+		w.Unlock()
 		return nil
 	}
 
+	// Forward the data to the subscription channel.
+	// We release the lock while calling the forward function to avoid holding
+	// the lock while doing potentially slow user code.
+	w.Unlock()
```
csilvers, line 170: In this case, I do agree that having to remember to unlock in all the exit paths is fragile.  It may make sense to pull all this code out into its own function so you can use `defer` to close.

graphql/websocket.go
```diff
@@ -247,15 +266,19 @@ func (w *webSocketClient) Unsubscribe(subscriptionID string) error {
...
 	}
-	err = w.subscriptions.Unsubscribe(subscriptionID)
-	if err != nil {
-		return err
-	}
+
+	w.Lock()
+	defer w.Unlock()
+	w.subscriptions.closeChannel(subscriptionID)
+	w.subscriptions.delete(subscriptionID)
```
csilvers, line 273: The old `Unsubscribe` didn't delete, if I'm reading the code right.  Is this a bugfix?

csilvers (commented):
I'm not really qualified to review, but I appreciate being explicit about what functions need the lock held!

benjaminjkraft:
I think this is now replaced by #417 + #418, thanks Mark!

PR #413, opened 2026-01-29
fix: add omitempty support for list types with pointer_omitempty

gofulljs:
fixed `pointer_omitempty` not being applied to list types when `use_struct_references` is enabled. List fields like `[String!]` now correctly get the `omitempty` JSON tag


benjaminjkraft (approved):
seems like a bug to me, thanks for the fix!

benjaminjkraft:
I think the test failure is still a flake, but it looks like we don't have a CLA from you, could you fill that out?

gofulljs:
I’ve signed the CLA, thanks.

benjaminjkraft:
Oops, sorry for the delay, merging now!

PR #417, opened 2026-05-20
Fix deadloack introduced in #402

dnerdy:
This PR fixes the deadlock introduced in #402. The deadlock was occurring because the forwardWebSocketData function was holding a lock on the subscription map (in the listenWebSocket goroutine) while Unsubscribe was blocked attempting to aquire the same lock. The forwardWebSocketData function was also blocked attempting to send data on the interfaceChan channel, which had no readers in the test being run.

After fixing the locking (by making sure the lock isn't held when forwardDataFunc is called), a race condition cropped up between sending on interfaceChan (in forwardWebSocketData) and closing interfaceChan in Unsubscribe. This was fixed by making the listenWebSocket goroutine the "owner" of the interfaceChan channel -- it is now the only goroutine that sends on and closes the channel. Other goroutines singal interfaceChan should be closed by setting `_hasBeenUnsubscribed` on the subscription, which is protected by a short-held lock.

Note that there's a similar data race involving isClosing on webSocketClient. isClosing is set when Close is called (while holding a lock) but read in listenWebSocket without a lock. #418 fixes this race (so the `-race` flag can be added when running tests).

I have:


graphql/subscription.go
```diff
@@ -13,10 +12,40 @@ type subscriptionMap struct {
...
+	// and closing of the channel, so is no possibility of races between send
+	// and close.
 	interfaceChan       interface{}
+
 	forwardDataFunc     ForwardDataFunction
 	id                  string
-	hasBeenUnsubscribed bool
+
+	// Hold when accessing _hasBeenUnsubscribed
+	hasBeenUnsubscribedMu sync.Mutex
```
csilvers, line 33: I think the copylocks (below) is a real problem -- you have to make the mutex a pointer and then hae a constructor for this struct.

csilvers (commented):
Thank you for taking this on!  I'll leave it to ben to review, my head always gets fuzzy thinking about locking.

graphql/subscription.go
dnerdy: I've updated the code to use *subscription in the subscription map, which should work as well.

benjaminjkraft (approved):
Thanks!

HaraldNordgren:
Thanks a lot, @dnerdy!

PR #418, opened 2026-05-20
Fix isClosing data race

dnerdy:
This PR fixes a data race related to the webSocketClient isClosing field. Before this change the field was set in the Close method while holding a lock but read in the listenWebSocket method without holding a lock. The point of the isClosing flag was to prevent sending on errChan after close. Instead of having a hidden dependecy between a flag and the errChan channel state, this PR replaces isClosing with a flag named exitListenWebSocket, and when the flag is set the listenWebSocket closes errChan itself. listenWebSocket is the only goroutine that writes to errChan, so there's no longer any possibility of writing on a closed channel -- the listenWebSocket goroutine now effectively "owns" errChan.

I have:


graphql/websocket.go
```diff
@@ -44,15 +44,19 @@ const (
 )
 
 type webSocketClient struct {
-	Dialer        Dialer
-	header        http.Header
-	endpoint      string
-	conn          WSConn
-	connParams    map[string]interface{}
+	Dialer     Dialer
+	conn       WSConn
```
dnerdy, line 48: Struct fields were reordered due to the following lint error:
```
graphql/websocket.go:46:22: fieldalignment: struct with 104 pointer bytes could be 80 (govet)
```

graphql/websocket.go
```diff
@@ -126,17 +122,20 @@ func (w *webSocketClient) listenWebSocket() {
 				sub.interfaceChan = nil
 			}
 		})
-		if w.isClosing {
+		w.exitListenWebSocketMu.Lock()
+		if w.exitListenWebSocket {
+			close(w.errChan)
 			return
```
benjaminjkraft, line 129: Do we need to unlock here?

benjaminjkraft (approved):
Nice, thanks for tracking this down!

graphql/websocket.go
dnerdy: Yes, indeed! Fixed.

PR #419, opened 2026-05-25
Add `omit_unreferenced_implementations` config option (#416)

### LINKED ISSUE #416: Interfaces with many implementations generate massive output even when only one type is queried via fragment

When a query selects from an interface type that has many implementations in the schema (e.g. `Node` in the GitHub API with 300+ types), genqlient generates a struct and `implementsGraphQLInterface` marker method for every single implementation — even if the query only uses fragments for one or two of them.

For example, this query:

```graphql
query GetThread($threadID: ID!) {
    node(id: $threadID) {
        ... on PullRequestReviewThread {
            comments(first: 100) {
                nodes { body }
            }
        }
    }
}
```

generates thousands of lines like:

```go
    func (v *GetThreadNodeOrganization) implementsGraphQLInterfaceGetThreadNode()            {}
    func (v *GetThreadNodeOrganizationInvitation) implementsGraphQLInterfaceGetThreadNode()  {}
    func (v *GetThreadNodePackage) implementsGraphQLInterfaceGetThreadNode()                 {}
    func (v *GetThreadNodePackageFile) implementsGraphQLInterfaceGetThreadNode()             {}
    // ... hundreds more
```

Each unreferenced type gets an identical struct (only shared fields) plus getter methods, marshal/unmarshal cases, and interface marker methods. The only type that actually has unique fields is `PullRequestReviewThread`.

Only generate full implementation structs for types explicitly referenced by inline fragments or named fragment spreads.

Unreferenced types could share a single catch-all "other" struct with just the
shared fields.

**Describe alternatives you've considered**

- `# @genqlient(struct: true)` works when there are no fragments, but errors when we have fragments.
- `# @genqlient(bind: "MyCustomType")` works too, but I'm using genqlient because I can't be trusted to manually maintain types.

**Additional context**

I have an llm-generated branch that addresses this. If you are interested in pursuing this issue I can use that as a starting point for implementation.

WillAbides:
Closes #416

Adds an opt-in `omit_unreferenced_implementations` config option that collapses unfragmented interface/union implementations into a single catch-all struct per interface selection.

With `omit_unreferenced_implementations: true`:

- Per-type structs are generated only for implementations actually referenced via an inline or named fragment in the selection set.
- A single catch-all struct `<…>GenqlientOther` is generated for each interface/union selection. It carries the interface's shared fields and satisfies the generated Go interface like any other impl. I chose the "GenqlientOther" suffix because it is very unlikely to be used in a real query.
- Defaults to `false` for backward compatibility.

### Tests

- Snapshot tests for the inline-fragment, named-fragment-on-interface, named-fragment-on-concrete-type, union, abstract-fragment-spread-inside-concrete-inline-fragment, and all-impls-collapse cases. The `Build` subtest confirms each snapshot compiles.
- Integration test in `internal/integration/omitimpls/` that exercises both the typed-impl path and the catch-all path end-to-end against the existing gqlgen test server.

I have:
- [x] Included a link to the issue fixed, if applicable (#416)


WillAbides:
I tried this branch with the code that originally made me want this feature, and I found a few things to address:
- `omit_unreferenced_implementations` should be settable with a directive
- A `catch_all_suffix` config would be nice
- Similar to above, a `catch_all_name` directive

PR #420, opened 2026-06-16
Add global "flatten" option to genqlient.yaml

HaraldNordgren:
Fixes #404.

Adds a global `flatten` option to `genqlient.yaml` that applies `@genqlient(flatten: true)` to every operation and named fragment, so flattenable fragment-spreads are flattened project-wide without adding the directive per query. A local `@genqlient(flatten: ...)` still wins, and it's only applied where flattening is valid, so it's safe to enable globally.


HaraldNordgren:
Ping @benjaminjkraft

benjaminjkraft:
Seems reasonable, thanks!

PR #422, opened 2026-07-17
chore(deps): replace archived gopkg.in/yaml.v2 with go.yaml.in/yaml/v3

levenleven:
`gopkg.in/yaml.v2` is archived. This swaps genqlient's config parsing to the maintained `go.yaml.in/yaml/v3` (the YAML organization's successor to `gopkg.in/yaml.v3`).

The migration is behavior-preserving vs yaml.v2, verified by comparing both libraries on genqlient's config shape:
- YAML 1.1 booleans (`yes`/`no`/`on`/`off`) still decode to bools
- octal-looking strings (e.g. `0777`) stay strings, not coerced to ints
- unknown fields are still rejected (`KnownFields(true)` replaces `UnmarshalStrict`)
- empty/comment-only config files still parse to an empty config
- `StringList`'s obsolete `UnmarshalYAML` is unchanged (v3 still supports it)

A new `testdata/validConfig/YAML11Booleans.yaml` fixture locks in the boolean and octal-string parity.

Note: the transitive `gopkg.in/yaml.v3` remains, pulled in only by test-only deps (gqlgen, gqlparser, testify); it will drop once those upstreams migrate.


generate/config.go
```diff
@@ -322,8 +325,12 @@ func ReadAndValidateConfig(filename string) (*Config, error) {
...
 
 	var config Config
-	err = yaml.UnmarshalStrict(text, &config)
-	if err != nil {
+	// yaml.v3 has no UnmarshalStrict; a decoder with KnownFields(true) is the
+	// equivalent, rejecting unknown fields. An empty file decodes to io.EOF,
+	// which we treat as an empty config (as yaml.v2's UnmarshalStrict did).
+	dec := yaml.NewDecoder(bytes.NewReader(text))
+	dec.KnownFields(true)
+	if err = dec.Decode(&config); err != nil && !errors.Is(err, io.EOF) {
```
ccoVeille: I'm not fan of the error inlined in the if

other checks on the error are not inlined

This would be more readable

```suggestion
	err = dec.Decode(&config)
	if err != nil && !errors.Is(err, io.EOF) {
```

ccoVeille (approved):
yaml/go-yaml maintainer here

Thanks for migrating to the maintained lib.

One minor remark, otherwise LGTM 👍

generate/config.go
levenleven: ✅

generate/config.go
```diff
@@ -322,8 +325,13 @@ func ReadAndValidateConfig(filename string) (*Config, error) {
 	}
 
 	var config Config
-	err = yaml.UnmarshalStrict(text, &config)
-	if err != nil {
+	// yaml.v3 has no UnmarshalStrict; a decoder with KnownFields(true) is the
+	// equivalent, rejecting unknown fields. An empty file decodes to io.EOF,
+	// which we treat as an empty config (as yaml.v2's UnmarshalStrict did).
```
benjaminjkraft:
```suggestion
```
(helpful for review but not needed going forward)

benjaminjkraft (commented):
As a check that the package is legit, it's used by e.g. dd-agent: https://github.com/DataDog/datadog-agent/blob/pkg/config/schema/v0.82.0/pkg/config/schema/go.mod. Beyond that, should be good if tests pass.

benjaminjkraft (approved):

generate/testdata/snapshots/TestValidConfigs-YAML11Booleans.yaml
```diff
@@ -0,0 +1,26 @@
...
+  PackageBindings: ([]*generate.PackageBinding) <nil>,
+  Casing: (generate.Casing) {
+    Default: (generate.CasingAlgorithm) "",
+    AllEnums: (generate.CasingAlgorithm) "",
+    Enums: (map[string]generate.CasingAlgorithm) <nil>
+  },
+  Optional: (string) "",
+  OptionalGenericType: (string) "",
+  StructReferences: (bool) true,
+  Extensions: (bool) true,
```
benjaminjkraft, line 23:
```suggestion
  Extensions: (bool) true,
  Flatten: (bool) false,
```
(from merge with main, I think)

